PR #28, opened 2025-02-10
spatial teleportation SPELL

ATT2CHTranslationTeam:
This spell can create two points, α and β, by consuming dahal, and can teleport by consuming dahal and ESC. It features a teleportation guide and has restrictions that prevent teleportation during boss fights or conversations. The maximum teleportation distance is 140 meters.

I wrote a short plot for the acquisition method to make its appearance more reasonable, but it hasn't been organized yet.


ATT2CHTranslationTeam:
It seems that I have made a mistake，sorry.The following part is the indication that the effect of the potion is disappearing.
Players can control its switch through the options in the Book of Consciousness.

orian34:
The spell seems interesting, though it does make me a bit afraid. The teleportation mechanic would be prone to break some mechanics or layouts since they weren't made with this possibility in mind.

ATT2CHTranslationTeam:
This spell has many limitations. Firstly, when using teleportation, there is a 4-second casting time. If the distance from the casting location exceeds 4 meters during this period, the teleportation casting will be interrupted. Additionally, players cannot use teleportation while fighting bosses, and it is also impossible to cast spells during main and side story cutscenes. Moreover, teleportation cannot be used in the nightmare run scenario. I think the restrictions I have imposed are already quite substantial.

Piccomaster:
Indeed we never wanted to integrate a teleportation spell because the player can break a lot of mechanisms by teleporting through walls. Unless the spell prevents that?

PR #31, opened 2025-02-11
Tp spell

ATT2CHTranslationTeam:
This spell has many limitations. Firstly, when using teleportation, there is a 4-second casting time. If the distance from the casting location exceeds 4 meters during this period, the teleportation casting will be interrupted. Additionally, players cannot use teleportation while fighting bosses, and it is also impossible to cast spells during main and side story cutscenes. Moreover, teleportation cannot be used in the scenarios of "Nightmare Run" and "Ithax".


orian34:
Excluding ithax solves a number of worries I had.

ATT2CHTranslationTeam:
After reading Piccomaster's reply, this is my interpretation.
This spell requires the player to manually set waypoints for teleportation, and they can only teleport to places they have previously visited. In essence, this spell acts as a portable Waypoint.
It does not have the effect of forcing movement in a certain direction; it only allows teleportation between different predetermined points.
For example, when I'm fighting the Somniophages boss in Ouran, I have to fly a long distance each time to continue. However, with this spell, I can create a teleportation point in front of the boss. After defeating the boss, I can directly teleport back to the same spot in front of the boss. This saves a lot of time and avoids the lengthy and boring process.

orian34:
That sounds reasonable. Sounds worth a shot to me.

Piccomaster:
Indeed it is very good work thank you, it seems quite reasonable how the spell is made. But the teleports in front of the bosses without limitation of use and time seems a little too powerful to me.
I think that you should explain to me a little more its limitations that you have planned at the base, the Dahäl, the prices, its cost for each use? And depending on the levels what does it change, what is the difference between lvl1 and lvl5?

I think I will accept the pull request, but I will have to test and study the spell on my side to know how to integrate it into the game.

There are few bugs too :
- In [adventquest_functions/data/att2/functions/gameplay/dahal/action/spell32/obtain.mcfunction] why there is path to "/function att2:gameplay/dahal/action/spell100/go" in the commands book?

ATT2CHTranslationTeam:
I have concocted a brief stroy about the TP spells. It goes like this: After completing the main storyline of Angband, I obtained a "Lost Energy Space Gem" in etotsira(old). Then,Use the runes "VON ORG "Lost Energy Space Gem" AVE YOG"to synthesize "Space Gem with a Pinch of Energy Restored"  Afterwards, I acquired the TP spell there in etotsira(old).
spell cost:
TP COST:
lvl1->The farthest is 40 meters 300dahal 
lvl2->The farthest is 60 meters 300dahal
lvl3->The farthest is 80 meters 300dahal
lvl4->The farthest is 100 meters 300dahal
lvl5->The farthest is 140 meters 300dahal
The ESC consumption is determined based on the distance between the player and the teleportation point
0..60meters 1ESC
80-100meters 3ESC
100-140meters 5ESC
Set and tp Cooldown:
lvl1 8s
lvl2 7s
lvl3 6s
lvl4 5s
lvl5 4s
The cooldown time is probably a bit too short, but it can be longer. I think at least it shouldn't exceed 1 minute.
Set waypoint cost:
From level one to level five, 350 dahal points and one ESC are needed.

Regarding that bug, it can be left alone. I commented out those two lines and forgot to delete them.

ATT2CHTranslationTeam:
As my coding skills are not that great, some of the code may appear a bit redundant or strange. Also since whether players can TP to the teleportation point depends on whether the chunk where the teleportation point is located is loaded, which is related to the player's view distance, I didn't set the maximum teleportation distance too high. If there is a way to keep the chunk where the teleportation point is located always loaded, the maximum teleportation distance can be set even farther.

Piccomaster:
Ok that seems coherent to me, except that maybe I will change the means of obtaining the spell. And it will require a lot of work for the translation.
But why in this pull request of the TP spell there is the horses system ? Can you delete this change from this pull request ?

ATT2CHTranslationTeam:
I'm not very familiar with this system. I'm sorry for accidentally including the horse system.I'll upload a new PR

orian34:
> I'm not very familiar with this system. I'm sorry for accidentally including the horse system.I'll upload a new PR

That is because you can only have one PR at a time from a branch as it looks at all your commits from that branch.

If you want to keep making commits unrelated to a PR, I suggest you don't push your commits to your branch, keeping them as local edits. Then once the PR is done, you can send the extra commits as it won't be part of the current PR.

Piccomaster:
Ok very well I merged the last pull requests with the TP and horse system. I hope everything works perfectly, I will have to do some tests and modify some things like the translation so it will require more work before an update but I think it should do it. If it does not work well I will just make these additions inaccessible in the code, after if you have any corrections to make you can always do it of course.
Thank you, and can you tell me what pseudonym I could put in the credits for the help you gave us?

ATT2CHTranslationTeam:
My pseudonym is zhubi

PR #34, opened 2025-02-12
Questions about horses

ATT2CHTranslationTeam:
I think the horse in att2 is very awkward. First, it can only be used in the sylberland, and it's impossible to ride a horse in other dimensions. Moreover, in the sylberland, due to the mechanism of the horse, players have to repeatedly use the horse whistle to mark the location of the horse and control its movement. Every time they use the horse, they need to turn the horse whistle on and off repeatedly. If they are a little careless, the horse may run away somewhere unknown. Or if players use the horse whistle and then go somewhere else, if they want to use the horse, they have to go back to the place where they left the horse to get it. I find these processes very torturous and boring. I made a modification that changes the function of the horse whistle to storing and summoning horses. It teleports horses near the player to the spawn point's always-loaded chunk and can only store one horse. Horses can be summoned in all dimensions. At the same time, summoning horses also has restrictions similar to those of the tp spell, such as not being able to summon horses in boss rooms, nightmare parkour, ihtax, or during main and side story animation scenes


ATT2CHTranslationTeam:
Of course, if you don't want it to be recallable in all dimensions, just add some restrictions to make it only recallable in Sylberland
This function, like the horse whistle, is carried out in the consciousness book

Piccomaster:
Hmm ok interesting I quite agree, but I have a question. The player can buy as many horses as he wants from the merchants, so how does the system select the horse the player will summon?

ATT2CHTranslationTeam:
This function allows players to teleport the horse closest to them to the spawn point, and then teleport the horse from the spawn point back to their side.So the summoned horse is the one that was teleported to the spawn point

PR #38, opened 2025-02-14
Achievement branch fix AND nofight dialog loop or player cant leave

ATT2CHTranslationTeam:
When the player selects "nofight" and obtains the time gem, a red stone block is placed in the spawn area after detecting the time gem in the player's inventory. Then, the ouran_3_go process begins. However, at this time, ouran_2_go is still running and executing the count from 1502 to 1900. The concurrent operation of these two goroutines causes conflicts, preventing the player from teleporting properly and triggering an endless loop of dialogues.

SO When "nofight" is selected, add an instruction to clear the red stone block of ouran_2_go, and ouran_3_go will operate normally.


ATT2CHTranslationTeam:
About the issue of pseudonym：My pseudonym is zhubi

PR #42, opened 2025-02-16
Modify the spawn location of shulker boxes

ATT2CHTranslationTeam:
With the help of my friend "THEDAWNNEARDUSK", we have modified the spawn locations of both the Rune Ore and the checkpoints. Now, the Shulker Boxes that display ores will spawn normally on the blocks with attached ores, and the Shulker Boxes that indicate spawn points will correctly appear on the Redstone Ore.


ATT2CHTranslationTeam:
This is about the modification of the secret seeker spell
and About the issue of pseudonym：My pseudonym is zhubi

ATT2CHTranslationTeam:
If the shulker are generated at the previous positions, the ones that indicate ore will appear below the ore. This causes the shulker to be exposed and vulnerable to player attacks, which also looks quite bad. Moreover, the previous shulker that indicated respawn points would generate at the respawn locations instead of on the "redstone ore" that triggered the respawn points.

orian34:
Sounds good to me.

PR #43, opened 2025-02-16
Achievement branch fix and nofight dialog loop or player cant leave and  GPS location fix

ATT2CHTranslationTeam:
The GPS guidance for some gems does not correspond to the appropriate gem slots.

When the player selects "nofight" and obtains the time gem, a red stone block is placed in the spawn area after detecting the time gem in the player's inventory. Then, the ouran_3_go process begins. However, at this time, ouran_2_go is still running and executing the count from 1502 to 1900. The concurrent operation of these two goroutines causes conflicts, preventing the player from teleporting properly and triggering an endless loop of dialogues.

SO When "nofight" is selected, add an instruction to clear the red stone block of ouran_2_go, and ouran_3_go will operate normally.

### DIFF
--- a/adventquest_advancements/data/att2/advancements/collection/armor_iron/all.json
+++ b/adventquest_advancements/data/att2/advancements/collection/armor_iron/all.json
@@ -7,7 +7,7 @@
         "description": {"translate":"Find all Traditional armor"},
         "frame": "challenge"
     },
-    "parent": "att2:collection/armor_diamond",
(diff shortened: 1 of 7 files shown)


Piccomaster:
Are you really sure, did you test the dialogue before, the fight and the dialogue after Ouran again to see if it still works perfectly? This system is delicate and if it is touched without being tested it is sure that it risks causing major blocking bugs and so I want to be sure that it has been tested before accepting the pull request.

PR #46, opened 2025-02-17
Movable stock

ATT2CHTranslationTeam:
I replaced the original "stock" spell with a new "mobile stock" spell. Here's the principle of this spell: By combining data packs and resource packs, specific chest minecarts are made invisible. The original mule is replaced with an invisible chest minecart, increasing the storage space from 15 to 27. Additionally, a pet follow system is added to make it move with the player. The chest minecart itself is also given the original "stock" particle effects to make it look better. Moreover, when the player picks up the "recycle_armor_stand" from the recycle merchant, the chest minecart will teleport to the "recycle_armor_stand". When the player leaves, the chest minecart will teleport back to the player's side. And only one chest minecart can exist at a time

If the player does not want the chest_minecart to be nearby, they can use the "mind" book or the "stock spell book" and choose the "stock_in" option to transfer (make it disappear)

The function att2:gameplay/dahal/action/spell20/stock_in is the way to execute this command. If necessary, it can be added to the spell book

Since this function needs to be combined with a resource pack, if only the data pack is imported, the chest_minecart cannot be invisible


Piccomaster:
So as said in the previous message, it poses a problem because there are already minecart chests on the map it seems to me so they will be invisible if they work by resource pack.
In addition it implies too many changes I can not merge like that on the current develop branch especially if I merge for the next update the players will lose all their stock so it would be especially important that people already on the map do the update.

Then, have you tested the system, everything works even in the dimensions? There is no bugs, no chance to lose items in stock ? Because I do not see how it is possible to make this system work without having a real storage chest permanently loaded in the center of the map therefore without modifying the main chunk apart from the datapack.

Also, have you tested with several players? Is there no conflict between the minecart chests? Can we summon several at the same time of different levels without it being a problem?

Finally it is true that if on paper it seems really good, I want to be sure that everything works perfectly if I have to merge it especially on the same current branch. So if you can reassure me on all these points that would be great.

ATT2CHTranslationTeam:
As I said before, this system will only make minecarts with the tag "invminecarts.invisible_minecart" invisible. Other chests and minecarts on the map will not be invisible, and players will not lose the items in their stock before updating this system. This is because I am using the logic of the original system, and the items in "chest_minecart" are stored and replaced according to the chests you previously placed in the spawn point chunk, so there is no need to worry about losing the items in the original stock.

Also, this system will limit each player to have only one "chest_minecart" at the same time, that is, each player can only summon one "chest_minecart" following them. As for whether multiple players can use it simultaneously, we are still testing it, but at least so far, there have been no problems.

Piccomaster:
Ok thank you very much for your quick response, I understand better now. I will continue to think before accept the pull request definitively.

ATT2CHTranslationTeam:
I'm sorry to have added more work for you. I have to apologize here.I didn't conduct multiplayer testing for tp spell. I have only recently started testing these two spells, and just now, during the multiplayer use of tp spell, I discovered some issues. However, I will fix them soon. If you encounter any problems while testing my work, please let me know, and I will address them as quickly as possible.

Piccomaster:
Yes there is no problem I plan to release the update immediately. If there are corrections to be made, you can make other PRs, that suits me indeed it requires work on my side too, I thought I was done with the additions ^^. But thank you because these additions still make the map even better than it already is.

ATT2CHTranslationTeam:
It is precisely because you have created such a perfect map that I can contribute my passion to it. But there is no need to rush to release the new version. I am still testing some bugs, and I also have some small ideas that have not been implemented yet.

Piccomaster:
Very well in any case I looked at the other PRs and it suits me, I will merge all of them. And yes, I will take my time to correct all of that.

PR #47, opened 2025-02-17
adv format fixes + chest access

orian34:
also fixed all epic collection being broken


ATT2CHTranslationTeam:
The method of opening these chest through the "other" way is too fine, and I feel that few people will discover them.But i like it

orian34:
> The method of opening these chest through the "other" way is too fine, and I feel that few people will discover them.But i like it

Yeah those are mostly safety locks, to ensure there is a way to get them if you really want to.

PR #50, opened 2025-02-18
Pet not follow player bug fix (really!!)

ATT2CHTranslationTeam:
I'm really really really sorry. I was misled by some minor issues during testing about the pet disappearance problem. In fact, the reason why bats and phantoms can't normally follow players after teleportation is that there is no "PersistenceRequired" tag attached. According to the mechanics of the original Minecraft, both bats and phantoms will naturally despawn if they are more than 128 blocks away from players, which is the real reason why they can't follow players normally. Besides, I added English names to the pets to make them look better.


orian34:
By the way, if I can suggest something, I noticed you push every change separately, but you can edit multiple files at the same time, then making it a bigger commit. 

Otherwise it creates a lot of small commits taking the feed.

orian34:
I realized that the reason you have to do that is that you're working on the browser github, so you have to edit files individually.

If you are able to install it, I highly recommend you setup github desktop, it is very helpful and allows easier editing.

ATT2CHTranslationTeam:
Thank you for your suggestion. I'll try it as best as I can.❤

PR #58, opened 2025-02-21
fix stock/tp/secret seeker spell error +fix "show_particles:0" bug 

ATT2CHTranslationTeam:


orian34:
I noticed the nojelanth present 2 checkpoint was removed from detection?

ATT2CHTranslationTeam:
Of course, since this checkpoint is redundant

ATT2CHTranslationTeam:
I have a question: Why does using data merge to modify the potion effects of an entity seem to prevent the corresponding effects from triggering properly, and even though show_particles is set to 0, particle effects still appear in the end?

however, using the effect command to apply potion effects to monsters triggers them correctly.

Piccomaster:
I don't know why, but you are sure for example that function works on mobs ? /effect give @s minecraft:resistance infinite 1 true

PR #64, opened 2025-02-23
Issues of the mythique equipment

ATT2CHTranslationTeam:
Regarding the issue of the mythique equipment, it is indeed too weak; Perhaps, apart from being visually appealing, it has no other advantages. Although I have developed a system that allows it to level up through runes, I understand that it might be quite challenging for you to accept. I hope they can receive better enhancements making them far from useless.

I've compiled some fairly decent enhancements

Rukyrion's new effect: Deals damage to nearby enemies based on the player's current strength points.

Kinuil's new effect: Deals random damage to nearby enemies.

Warlord's new effect: Arrows shot will cause chain damage to nearby enemies.

Bloodeater: Improved the conversion efficiency of damage dealt (in fact, even if a player has a strength buff of level 256, they can only deal a maximum of about 800 damage, so the threshold should be reduced from 300 to 200; otherwise, the level 4 effect will never trigger).


Piccomaster:
So it's interesting indeed, the principle of mythical weapons at the base is that they are already the only indestructible weapons. It is really that some are not very powerful but they are not impossible to obtain and you must not fall on the other side by making mythical weapons too powerful. For example Rukyrion seemed quite powerful to me already.

However I see that the balancing seems correct to me for others, I will have to test but I think I will accept these changes they seem interesting to me.

In this case I even suggest finding a way to increase the power of the mythical weapon "Lone shadow" a little.

ATT2CHTranslationTeam:
Actually, I think Rukyrion could also incorporate some negative effects when the player uses it(For example, at level 4, when dealing damage to enemies, it could simultaneously reduce the player's own health or something similar...), much like Blindshield or Bloodeater, which might make it seem less "over powerful."

ATT2CHTranslationTeam:
I've conceived a new effect for "Lone Shadow": This state has a 6-second cooldown. This means that after triggering the "shadow buff" once, the player must wait 6 seconds before they can trigger the "shadow buff" again. After attacking an enemy, it inflicts a high-level slowness effect on nearby enemies for 2 seconds and applies a "shadow buff"(4 second)  to the player. During this time, the player becomes invisible (with black particle effects) and gains a darkness buff and a significant speed buff, continuously consuming Dahal for 4 seconds. It also accumulates the damage dealt to enemies during this period and inflicts corresponding damage to nearby enemies based on the accumulated damage, accompanied by a slow shadow sound effect.

For example: I attack an enemy and then gain the "shadow buff." During the four seconds that the "shadow buff" is active, I deal 2000 points of damage to the enemy. Then, after the "shadow buff" ends, I deal 500 points of damage to nearby enemies and apply a 5-second slowness effect to them.

Piccomaster:
Indeed it seems interesting as a mechanic although 500 damage points seems excessive, but in any case yes it is a good idea. On the other hand a little complex to code but it can be done.

ATT2CHTranslationTeam:
Similar to the previous description: After attacking an enemy, the player gains a 4-second "shadow buff," increasing speed by 10 points, 4 seconds of invisibility, and 5 seconds of darkness effect, while starting a 10-second cooldown. Nearby enemies are also inflicted with Slowness II for 4 seconds. During the "shadow buff," the total damage dealt by the player is accumulated and settled after 4 seconds. If the player switches to another weapon or if the Dahal is insufficient to maintain the "shadow buff," the damage settlement will occur prematurely.

When the "shadow buff" cooldown is 0, the player will be given particle effects while holding "Lone Shadow" for easier observation; when the "shadow buff" cooldown is greater than 1, there will be no particle effects.

Piccomaster:
Oh that's great, honestly that seems super cool. I'm going to merge the PR, and I'll test all that.

PR #65, opened 2025-02-23
fix stock/wolf bug

ATT2CHTranslationTeam:


Piccomaster:
"Stock cannot be used normally in non-overworld dimensions" What do you mean by that ? The stock can no longer be used in the Nether so in Angband and Billgart dimension ? Why did you change this line and add "execute in minecraft:overworld run " to "function att2:gameplay/dahal/action/spell20/copy_chest_minecart_to_chest" ?

ATT2CHTranslationTeam:
It's nothing; it's already fixed. The main issue was not setting the dimension for recording chest information. When importing items from chests or minecarts into a chest, the coordinates for the Nether or End dimensions were being selected, which caused the information to not be saved properly.

Piccomaster:
Oh ok, thanks for the quick reply ^^.

PR #76, opened 2025-03-01
fix tp spell have a "b"

ATT2CHTranslationTeam:


Piccomaster:
Sorry to late and these are not the correct line. I'll do fix later.

PR #78, opened 2025-03-01
Prevent the generation of too many Shulkers

ATT2CHTranslationTeam:


Piccomaster:
It's not the right color of the spell now and I already fixed the items, the lifetime item is not necessary to fixed.

PR #80, opened 2025-03-03
Chest particle effect optimization.

ATT2CHTranslationTeam:
I have built two systems for optimizing chest particle effects.

One is an optimization of the previous continuous loop method---(Although in the latest version, you changed the chest particle effects in the overworld to trigger once per second, I feel that while this change has significantly improved optimization, it has also lost a lot of practicality), so I optimized the initial method. This structure allows each chest to trigger particle effects four times a second, making it look almost the same as it did initially.

The other method is to assign particle effects only to chests within a 3x3x6 range in front of the player's view.

I can't really say which optimization is better because they felt about the same during my testing. As for the detection range of the second method, it can't be too large. To be honest, this method is ideal, but to actually implement it, a lot of detection commands need to be listed. If the detection range is expanded, the lag would also be quite high.

scoreboard players set chest_color TIMECOUNTER 1 can enable the detection for the first method, and scoreboard players set chest_color TIMECOUNTER 10 can enable the detection for the second method.

You can test it specifically.

As for the issue with "predicate," it is a very powerful feature that can detect many things that cannot be detected through ordinary scoreboards, and it can also achieve the kind of NBT/tag detection, greatly enhancing game optimization.

For example, when using execute @e, if tags or NBT that have been added are used for detection, the system will generate more operations, leading to a certain degree of lag. Switching to predicate for detection can slightly alleviate such lag.


ATT2CHTranslationTeam:
The architecture of "eye_effect" has been re-optimized, and now it only runs a little over 2000 commands per second, with just over 100 commands per tick, making the lag almost negligible.

ATT2CHTranslationTeam:
I noticed that many detections in the map use armor stands (as do some of the functions I have been working on). However, I recently learned that armor stands have a "Marker" option. When set to 1, the armor stand becomes a mode that cannot interact with players and essentially disappears from the map (not counted as an entity), but data can still be assigned to it. This is very helpful for some detections, as it can reduce entity lag on one hand and eliminate the blocking of player attacks by armor stands in certain situations on the other.

Piccomaster:
Ok you teach me a lot of interesting things about Minecraft indeed. I'll test that, it's good work!

PR #83, opened 2025-03-05
issue with wolf

ATT2CHTranslationTeam:
It seems that as long as the tag contains "class," the monster value mechanism will take effect on this entity, and eventually, the wolf's customname will change to ◈.

And its base values will also be changed to the corresponding values


ATT2CHTranslationTeam:
I have already tested it specifically, and there are no issues.

The specific optimization effect is very obvious. Previously, the TPS here was basically all red, and sometimes it could even reach the top of the screen =_=.

Piccomaster:
Removing the CLASS from the wolf will prevent you from defining the power it will have in attack and life... It must be added manually in this case.

PR #84, opened 2025-03-06
[bad news]Fix magmacube not dying.

ATT2CHTranslationTeam:
This is bad news.... I forgot to add the kill detection for magma_cube.... Very, very sorry...

When setting marker1 for an armor stand, its name will be displayed at its feet...


Piccomaster:
Don't worry, as long as it's not a major bug it's fine ^^. It's normal to have bugs after such an update, I'll do an update anyway so it's time for that. And even if you forgot other additions, it will be for the next update.

PR #85, opened 2025-03-07
Spell cooldown display.

ATT2CHTranslationTeam:
I have created a system to display spell cooldowns.

After a player launch a spell, if the spell is still on cooldown, it will prompt the remaining cooldown time of the spell.

I have organized all the languages, but since some of the "ARABIC LANGUAGE" I really don't know how to copy, I have kept it as "XXXXXX". If necessary, it would be best to replace it with the correct description. This is the "ARABIC LANGUAGE" that I did not copy: spell2 spell4 spell6 spell8 spell21 spell22 spell23 spell27 spell28 spell30 spell32 spell34 spell44


ATT2CHTranslationTeam:
I just discovered during testing that the chests in the exploration point warehouse have loot for reg1/c3t6, but there is actually no corresponding loot table in the data pack.

Piccomaster:
Ah indeed, a very good idea.

PR #87, opened 2025-03-09
Fixed the bug of riding through portals + accidental death of stock +guardian kill detection

ATT2CHTranslationTeam:
There is an issue with the dimensional teleportation detection range. If a player rides a horse into a portal, due to the large collision volume of the horse, the position detection will loop and continuously teleport the horse, trapping the player on the horse and preventing movement. Other entities, such as pets or others, might also encounter this problem.

So replacing @e with @a should have no impact.

### DIFF
--- a/adventquest_advancements/data/att2_test/advancements/test_mobskilled/guardian.json
+++ b/adventquest_advancements/data/att2_test/advancements/test_mobskilled/guardian.json
@@ -4,20 +4,20 @@
             "item": "minecraft:target"
         },
         "title": {"translate":"test"},
-        "description": {"translate":"magma_cube"},
+        "description": {"translate":"guardian"},
         "hidden": true,
         "show_toast": false,
         "announce_to_chat": false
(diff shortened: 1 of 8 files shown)


ATT2CHTranslationTeam:
I have recently been working on a system that records the time it takes for players to kill a boss, from the moment they start fighting the boss until the moment the boss's rewards are settled.

We can give players some extra rewards based on the time it takes to kill the boss...

I feel that this reward is actually optional, but I think it's quite interesting to be able to record the time taken to kill a boss. As for the extra rewards and related achievements, we can consider them further.

I don't know if this project will interest you

Piccomaster:
Indeed I had planned to add for each boss an exploit to kill them in a certain time but I abandoned the idea because the player can change difficulty and it was necessary to make a time for each difficulty, and testing the time to kill each boss of the map for each difficulty was too long for me.

But I like the idea.

PR #95, opened 2025-03-10
English voice + Other Minor Things

ElGrenier:
Sorry for this push "fourre-tout", I still try to clearly understand how Github works, so in the meantime, here is every change i made..


ATT2CHTranslationTeam:
hi，
“I realized that the reason you have to do that is that you're working on the browser github, so you have to edit files individually.

If you are able to install it, I highly recommend you setup github desktop, it is very helpful and allows easier editing”

This is what "orian34" told me, and I think it should also be helpful to you ^_^

PR #96, opened 2025-03-11
Record of time taken to defeat the boss.

ATT2CHTranslationTeam:
To be honest, the text part displaying the time is not very well designed by me...


ATT2CHTranslationTeam:
Since some bosses cannot be repeated, I did not do the related timing, but if needed, I can add it.

ATT2CHTranslationTeam:
I tested the boss kill times on Nightmare difficulty.

Character data: Max level "Corruption," strength points 35, and resistance buff 5, using the golden shovel "Lelrïm tar." These conditions can be considered quite conservative. The times are generally close to the limit (except for the "golem" boss shot by Billgart's arrows).
 They can be used as a reference.

I forgot to test the Elevator, but its time is related to the iron golem, so it probably doesn't need specific testing.

ATT2CHTranslationTeam:
Oh, I also forgot to switch to English, emm,, but it should be understandable, at least the boss names are in English.

ATT2CHTranslationTeam:
I tested the boss kill times on Nightmare difficulty.

Character data: Max level "Corruption," strength points 35, and resistance buff 5, using the golden shovel "Lelrïm tar." These conditions can be considered quite conservative. The times are generally close to the limit (except for the "golem" boss shot by Billgart's arrows).
They can be used as a reference.

I forgot to test the Elevator, but its time is related to the iron golem, so it probably doesn't need specific testing.

ATT2CHTranslationTeam:
Oh, I also forgot to switch to English, emm,, but it should be understandable, at least the boss names are in English.

orian34:
> Since some bosses cannot be repeated, I did not do the related timing, but if needed, I can add it.

Yeah I think it's fine to not have it for non repeatable bosses. It wouldn't feel fun or fair.

ATT2CHTranslationTeam:
I also tested the bosses on Master difficulty, but I feel there isn't much difference from Nightmare difficulty; the defeat times for some bosses were only shortened by about 10 seconds.

ATT2CHTranslationTeam:
I noticed that the chest located at -4927 83 -5015 in Ryliath seems to have no way to get inside. Is there a need for some alternative method to make it openable?

ATT2CHTranslationTeam:
Master difficult test

Piccomaster:
Wow indeed, that's a lot of work. I'm fixing conflicts and merging everything. However, I'll have to test each boss to make sure they all work.

ATT2CHTranslationTeam:
Oh, I rushed the operation a bit, some scores are not fully corrected, and there are some errors in the timing information, but it's just minor issues

orian34:
> I noticed that the chest located at -4927 83 -5015 in Ryliath seems to have no way to get inside. Is there a need for some alternative method to make it openable?

Fixed that thanks.

PR #101, opened 2025-03-14
Main quest guidance (more detailed).

ATT2CHTranslationTeam:
Side quests haven't been started yet.


ATT2CHTranslationTeam:
Mainly, it uses if/unless to determine the current quest progress and provides guidance corresponding to the current stage.

Piccomaster:
Ah, okay, what you've done is a huge undertaking. Are you sure of your work? It will require testing all the quests to ensure there are no bugs.

But indeed, if it's a more specific indication, I'd be happy to accept these changes.

PR #103, opened 2025-03-14
The chunk where the teleportation point is located will be force-loaded

ATT2CHTranslationTeam:
This means that we can now increase the teleportation distance of the "Spatial Teleportation" spell....


ATT2CHTranslationTeam:
This force-loaded chunk will not conflict with the spawn chunk force-loaded by the map itself, so there's no need to worry about other issues arising.

Piccomaster:
I notice you change this command : `tp @e[type=minecraft:shulker,tag=MainObjective] ~ ~-200 ~`
I can advise you instead to put the height in absolute value to be sure that it falls at the same level without any problem that it is outside the map like this :
`tp @e[type=minecraft:shulker,tag=MainObjective] ~ -1 ~`

PR #104, opened 2025-03-14
Spell damage balance.

ATT2CHTranslationTeam:
This is the damage I modified in the Simplified Chinese version, which you can use as a reference.
 If you find it unacceptable, just close this PR.

### DIFF
--- a/adventquest_functions/data/att2/functions/gameplay/dahal/action/spell11/damage/lvl1/bonus0.mcfunction
+++ b/adventquest_functions/data/att2/functions/gameplay/dahal/action/spell11/damage/lvl1/bonus0.mcfunction
@@ -3,8 +3,8 @@
 #Fireline lvl1 effect											#
 #################################################################
 
-execute if score @p NUMEROJOUEUR matches 1 run damage @s 5 minecraft:player_attack by @p[scores={NUMEROJOUEUR=1}]
-execute if score @p NUMEROJOUEUR matches 2 run damage @s 5 minecraft:player_attack by @p[scores={NUMEROJOUEUR=2}]
-execute if score @p NUMEROJOUEUR matches 3 run damage @s 5 minecraft:player_attack by @p[scores={NUMEROJOUEUR=3}]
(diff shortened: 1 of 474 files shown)


Piccomaster:
Ok thanks I see but I can't accept this, it's to high damage, even power ray is 30% more powerfull than before the 1.0.0 version... 
All damage has been doubled, it's a huge buff even after my changes. Some have tripled their damage compared to the previous version.

ATT2CHTranslationTeam:
I'm not quite sure how you calculated the damage. In version 3.0.0, the damage of Power Ray was instant damage/healing 9, corresponding to a value of 1536, and it was magic damage, which did not calculate the damage reduction from armor. Now, this damage is player_attack, which does calculate the damage reduction from armor, with the maximum damage being only 1500, and this is after the mana recovery has been weakened. I think this damage is already quite conservative.

Piccomaster:
So maybe not 1.0.0 but we did some balancing later, juste after the 1.1.0 then, but the most stable we choosed was an effect "give @s minecraft:instant_damage 1 8", which makes max: 768 damage
And all that it's without take account of rune effect upgrade system...

ATT2CHTranslationTeam:
Emmm, actually, if the command is in this form: "minecraft:instant_health 1 8", it means instant healing of 9 over 1 second, not instant healing of 8.This level calculation is from 0-9, when it is 0, it corresponds to a buff level of 1.

ATT2CHTranslationTeam:
> Emmm, actually, if the command is in this form: "minecraft:instant_health 1 8", it means instant healing of 9 over 1 second, not instant healing of 8.This level calculation is from 0-9, when it is 0, it corresponds to a buff level of 1.

You can try typing “give @s luck 200 8” into the game, and eventually you'll find the luck level is 9 the status bar.

Piccomaster:
Oh my bad, OK, I get it, we miscalculated the conversion. We were calculating in number of hearts and not in number of half-hearts.
Wow, that's a mistake that dates back to the conversion to damage. Okay, I'll accept the pull request then.
It's crazy, I stupidly missed that detail, thanks ^^

PR #105, opened 2025-03-14
English VA Cinematic change

ElGrenier:
Changed the way cinematic are launched so each different voice acting has the right timing

Currently all "en" cinematic are the copy paste of the "fr" version, but it will be modified when actively implementing the voice inside the game

/!\ Please don't merge until I ask to, its here so you have access to the change, so please wait.


Piccomaster:
Oh shit, you know what that means, testing all the cutscenes in the game again ^^. Good luck, I'll wait, but you shouldn't wait too long or else make another pull request because there could be changes between the time you've finished and the update...

orian34:
Yeah, if you have changes like that that require long term, it's best to not make a PR yet, keeping it in the stash as you work on it.

PR #109, opened 2025-03-18
Fixed bugs in the rune synthesis system, added runicwords name display

ATT2CHTranslationTeam:
You need to input 'function att2:gameplay/runes/merge_runicstand" to return the position and movement of the armor stand to normal.

When containing "Mot Runic Incomplet," particle effects will be displayed to alert the player.

When the rune is not activated, it will emit red particle effects to alert the player.


ATT2CHTranslationTeam:
Emmm, it seems there are some issues; "function/gameplay/runes/recipes_test/rune_number/reduce" is completely useless and can just be deleted.

Piccomaster:
I can't merge, there is too conflit with this file.

PR #110, opened 2025-03-19
Fixed bugs in the rune synthesis system, added runicwords name display

ATT2CHTranslationTeam:
You need to input 'function att2:gameplay/runes/merge_runicstand" to return the position and movement of the armor stand to normal.

When containing "Mot Runic Incomplet," particle effects will be displayed to alert the player.

When the rune is not activated, it will emit red particle effects to alert the player.

Hope there are no issues.


ATT2CHTranslationTeam:
I found a small issue: when refreshing rune bonuses, all players' health is reset and fully restored, which could potentially be exploited to make players invincible under certain circumstances...


Moreover, I can't think of any particularly good solutions. If using "damage" to reset the player's instant health, it has a very undesirable characteristic: the health restored by instant healing would be calculated into the "damage", meaning that if "damage" and instant healing are run simultaneously, their values would be added and subtracted. However, if the damage exceeds the player's current health limit, the player would die....

Piccomaster:
There is a very simple trick to fix this, it is to restore life to all players only who are in Ryliath or in front of the rune system ^^ .

ATT2CHTranslationTeam:
Oh, indeed, but Ryliath is a bit too large; I think it could be reduced to just the entire rune room.

ATT2CHTranslationTeam:
Felroth's equipment enchantments are all Protection 10 across all difficulties, which seems a bit too strong for lower difficulties.

I'm not quite sure if it's a mistake, but when the total protection level is 20, it reaches the damage reduction cap of 80% for protection enchantments.

Piccomaster:
Ah indeed, great job, I'll try that.

PR #111, opened 2025-03-19
Add boss timing rewards

ATT2CHTranslationTeam:
Trying to add corresponding rewards, which might be a bit unbalanced.

Among them, "if score karon_m BOSS_TIME matches ..1"
 "..1" represents within 2 minutes.


Piccomaster:
Ah, that's a great idea ! On my side I just planned, for each boss killed only in MASTER before a certain time, to add an advancement with Exploit points gained in the "Boss Trophies" category ("hunting" folder).
The problem I have is that I don't know how long it takes to beat each boss to unlock the Exploit. If you have any idea of ​​the time, I'd love to hear it. Or at worst, I'd take the time from the rewards you've started to place as an example.

I think I'll take inspiration from your "#boss time 3" score, what do you think?

ATT2CHTranslationTeam:
> Ah, that's a great idea ! On my side I just planned, for each boss killed only in MASTER before a certain time, to add an advancement with Exploit points gained in the "Boss Trophies" category ("hunting" folder). The problem I have is that I don't know how long it takes to beat each boss to unlock the Exploit. If you have any idea of ​​the time, I'd love to hear it. Or at worst, I'd take the time from the rewards you've started to place as an example.
> 
> I think I'll take inspiration from your "#boss time 3" score, what do you think?

There is not much difference between the fastest time it took to kill the boss on master/nightmare difficulty and the "bosstime 3" I added to the "bosstime rewards", but considering the actual operation will be very difficult, I set the time in "bosstimerewards to be a bit more generous. In short, you just need to write the achievements according to the time of “bosstime 3”

PR #113, opened 2025-03-21
Casino tier/output equipment adjustment.

ATT2CHTranslationTeam:
To separately create a loot table and then replace some of the bats that generate dyes in "launchXXX," which would allow the production of equipment from other dimensions without affecting the points.


Piccomaster:
Ok but, the problem is that I can't accept if the player can obtain dimensional equipment before he has been able to access it through the main quest.

ATT2CHTranslationTeam:
> Ok but, the problem is that I can't accept if the player can obtain dimensional equipment before he has been able to access it through the main quest.

its ok， i have change it

Piccomaster:
Ok thanks it's great but I think you can take the moment of the main quest where the player first enters the dimension, no need for him to finish the entire dimension to unlock the bet.

PR #115, opened 2025-03-28
Reputation points and shop system.

ATT2CHTranslationTeam:
This system will increase the probability of refreshing high-quality equipment based on the player's current reputation points.


ATT2CHTranslationTeam:
Most of the repetitive copy-paste work was done using mouse and keyboard macros. I've also briefly tested the entire system, and there shouldn't be any major issues.

For pricing, I chose to retain the original prices, with some "leg" items slightly increased.

If price adjustments are needed, they should be made in "functions\gameplay\shop\purchase\price_set".

The discount system's range is 70% (400 reputation points) to 130% (-100 reputation points). Each time reputation points are increased or decreased, the item prices are updated once.

Piccomaster:
Great job once again, for the moment I have nothing to say, I will have to test all of this and see if it would be necessary to balance certain prices again.

PR #117, opened 2025-03-30
A new spell that continuously converts nearby equipment into money and xp

ATT2CHTranslationTeam:
It exists as a "pet" in this map.

Players can select items of specific qualities for conversion in the spell book, and there is a "fool-proof" mechanism—the spell only converts items with "pickupdelay:0". This means if important equipment is accidentally dropped, it can be picked up immediately without being instantly cleared.

The positioning of this spell is relatively late-game. My idea is that after players reach Black Market level 5, they can synthesize it using the "Recycle" spell and some rune materials.

Because in the late game, the cooldown of the "Recycle" spell and the limitations on clearing items often can't keep up with the amount of items generated, I created an advanced version of the "Recycle" spell. This allows players to explore without worrying about the tedious and cumbersome process of handling excess equipment.

And it also has a hidden mechanism: after consuming a certain number of items, this pet will continuously spew out large amounts of diamonds, small coins, and big coins.

And its Chrontons conversion rate has been adjusted, so even if the discount reaches its maximum, it won't create an infinite loop situation with it.


Piccomaster:
Wow, I wasn't expecting that... I'm a little worried it might make the recycling spell useless, right? Was it really necessary?

I haven't tested it yet, but does that mean all items on the ground are automatically converted into XP or Chronotons?

ATT2CHTranslationTeam:
It might make the previous "Recycle" spell less necessary, but this new spell mechanism is indeed stronger and more convenient than the "Recycle" spell, and I position it for use in the later stages of the game, and its upgrade requires consuming a large amount of equipment, and its chronotons and xp conversion efficiency is not high at low levels.

PR #119, opened 2025-03-31
Added some special effects for mob invasion and slightly increased its probability

ATT2CHTranslationTeam:
During testing, I found that the trigger probability for this mob invasion is really too low—it might not even trigger once in over an hour.

Slightly increased the spawn probability of mob invasion and added a guarantee mechanism to prevent it from failing to trigger consecutively.

### DIFF
--- a/adventquest_advancements/data/att2/advancements/treasure/sovereign/bundle.json
+++ b/adventquest_advancements/data/att2/advancements/treasure/sovereign/bundle.json
@@ -1,0 +1,26 @@
+{
+    "display": {
+        "icon": {
+            "item": "minecraft:bundle"
+        },
+        "title": {"translate":"Possibly useful, but usefulness is unlikely"},
+        "description": {"translate":"Obtain a small pouch"},
+        "frame": "goal"
+    },
+    "parent": "att2:treasure/sovereign",
+        "rewards": {
+            "experience": 777
+    },
+    "criteria": {
+        "bow_172": {
+            "trigger": "minecraft:inventory_changed",
+            "conditions": {
+                "items": [
+                    {
+                    "items": ["minecraft:bundle"]
+                    }
+                ]
+            }
+        }
+    }
+}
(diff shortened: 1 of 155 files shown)


Piccomaster:
Why did you do this :
`execute if score RNG100 RNG100_Variable matches ..49 run scoreboard players operation RNG100 RNG100_Variable += GM INVASION`

You should not touch these RNG variable, because it concerns other mechanisms. It does not work like that. Moreover, if you played with a low reputation rate and Exploit, it is normal that it is rare since it evolves with the TITLE_ID and therefore the mobs spawn rate increases the more the player advances in the game.
So no need to complicate things when you just have to change this number: 33..66, which indicates that you have a 1/3 chance of it happening.

I can't accept this pull request like this sorry.

ATT2CHTranslationTeam:
oh, true, I forgot to consider the other uses of this random number, but I really think the chance of generating a monster swarm is really low, although is a 33% chance of generating, due to the random selection of "choose time", its generation interval will be very long, which makes the number of times this appears very small

Piccomaster:
Very well, then, if you change these functions and just replace the correct numbers to increase the spawn chance, I accepte the pull request.

So remove this:
execute if score RNG100 RNG100_Variable matches 33..66 run function att2:gameplay/enveffect/mobs_infestation/summon_trigger_position
execute if score RNG100 RNG100_Variable matches ..49 run scoreboard players add GM INVASION 5
execute if score RNG100 RNG100_Variable matches ..49 run scoreboard players operation RNG100 RNG100_Variable += GM INVASION
execute if score RNG100 RNG100_Variable matches 50.. run function att2:gameplay/enveffect/mobs_infestation/summon_trigger_position
execute if score RNG100 RNG100_Variable matches 50.. run scoreboard players set GM INVASION 0

And change just like this:
executed if score RNG100 RNG100_Variable matches 1..50 run function att2:gameplay/enveffect/mobs_infestation/summon_trigger_position

ATT2CHTranslationTeam:
For ElixirVitae, I'm unsure if there's an error - its potion duration is tied to the 'SET' bonus timer rather than the standard potion ('PO') duration

PR #121, opened 2025-04-04
A brand-new spell-casting system that retains the original casting logic, utilizing the Dahal Focus for spell invocation.

ATT2CHTranslationTeam:
When the player holds a spellbook, pressing the "Q" key drops the book and switches to Focus Mode. While holding the Focus, right-clicking casts the spell.

When holding the Focus, pressing the "Q" key switches to Spellbook Mode, allowing you to select spells of different tiers.

While holding the Focus, press [Q] to enter Spellbook Mode and select a spell tier. The chosen tier remains active until manually changed – no need to reselect for every cast.

Spells like Nova/Corrupted/Tiid Klo Ul will have their sustained effects interrupted if another offensive spell is cast. However, they remain compatible with non-damaging spells – for example, during Nova's duration, players may freely use 'Secret seeker' or 'Agility Pet' without breaking the effect.


ATT2CHTranslationTeam:
i think this system will open up a whole new world of possibilities

ATT2CHTranslationTeam:
> When the player holds a spellbook, pressing the "Q" key drops the book and switches to Focus Mode. While holding the Focus, right-clicking casts the spell.
> 
> When holding the Focus, pressing the "Q" key switches to Spellbook Mode, allowing you to select spells of different tiers.
> 
> While holding the Focus, press [Q] to enter Spellbook Mode and select a spell tier. The chosen tier remains active until manually changed – no need to reselect for every cast.
> 
> Spells like Nova/Corrupted/Tiid Klo Ul will have their sustained effects interrupted if another offensive spell is cast. However, they remain compatible with non-damaging spells – for example, during Nova's duration, players may freely use 'Secret seeker' or 'Agility Pet' without breaking the effect.

orian34:
For some reason the PR deletes all the gameplay tips, that's a problem.

Also does that mean that all the spellbooks will be removed? I'm not sure if I like such a fundamental change, especially as having to throw an item with Q in combat is a very big problem. That's adding a significant risk of players throwing their weapons away by mistake.

ATT2CHTranslationTeam:
> i think this system will open up a whole new world of possibilities

To get it to work properly, you need to type "function att2:gameplay/dahal/launcher/initialize", then refresh the spellbook and it will work properly after selecting the spell level.

ATT2CHTranslationTeam:
> For some reason the PR deletes all the gameplay tips, that's a problem.
> 
> Also does that mean that all the spellbooks will be removed? I'm not sure if I like such a fundamental change, especially as having to throw an item with Q in combat is a very big problem. That's adding a significant risk of players throwing their weapons away by mistake.

I used AI to translate other language versions without modification, and as for using Q to switch in battle, this should not exist, because players only need to select spell level and then switch to the Dahal casting state, and only need to hold the corresponding casting tool in battle to cast spells, without performing other operations

orian34:
> I used AI to translate other language versions without modification, and as for using Q to switch in battle, this should not exist, because players only need to select spell level and then switch to the Dahal casting state, and only need to hold the corresponding casting tool in battle to cast spells, without performing other operations

How does a player change spells with this system? Also it's not too uncommon to use different spell tiers in combat, and it seems to be only working for a player that has all spells at the same level, otherwise it becomes weird.

ATT2CHTranslationTeam:
> > I used AI to translate other language versions without modification, and as for using Q to switch in battle, this should not exist, because players only need to select spell level and then switch to the Dahal casting state, and only need to hold the corresponding casting tool in battle to cast spells, without performing other operations
> 
> How does a player change spells with this system? Also it's not too uncommon to use different spell tiers in combat, and it seems to be only working for a player that has all spells at the same level, otherwise it becomes weird.

I'm not quite sure what you mean by that, players can choose different spell levels in the spellbook and will get the corresponding dahal casting device, not all spells can only use the highest level.

orian34:
So you mean it's an addition to the current system? As in you can keep using spellbooks like normal, or throw the catalyst for the system you added?

ATT2CHTranslationTeam:
> So you mean it's an addition to the current system? As in you can keep using spellbooks like normal, or throw the catalyst for the system you added?

It's about the same as what you said.
This system corresponds a casting device to each spellbook, players only need to select the spell level, then switch the casting device status to directly use the spell, is no need to repeatedly open the spellbook to select when using different spells.

Piccomaster:
I'm afraid I didn't understand everything correctly. I'd have to test it, but I'm afraid to accept a change as fundamental as the Dahäl for the map...

The Q key when you're on a French keyboard is to move left. I think you're talking about the key to throw items?

Actually, the problem is that at this level of the game, you absolutely mustn't change players' habits for the sake of a fundamental gameplay system.
Can you confirm that, for example, the basic gameplay of the Dahäl doesn't change at all?

Especially given the number of changes, it's almost certain that upgrading to the next version during the adventure would be difficult to implement...
I'm starting to think more and more about whether I should call the next update 5.0.0, hmm...

ATT2CHTranslationTeam:
> I'm afraid I didn't understand everything correctly. I'd have to test it, but I'm afraid to accept a change as fundamental as the Dahäl for the map...
> 
> The Q key when you're on a French keyboard is to move left. I think you're talking about the key to throw items?
> 
> Actually, the problem is that at this level of the game, you absolutely mustn't change players' habits for the sake of a fundamental gameplay system. Can you confirm that, for example, the basic gameplay of the Dahäl doesn't change at all?
> 
> Especially given the number of changes, it's almost certain that upgrading to the next version during the adventure would be difficult to implement... I'm starting to think more and more about whether I should call the next update 5.0.0, hmm...

The casting process of this system is the same as before, holding the dahal launcher and right-clicking to launch, except that it can quickly between the two forms of spellbook and dahal launcher through the "Q" key, and each spellbook has a separate dahal launchet. For example, players can 10 spells in the 0-9 quick slot, and then they only need to press the number key or scroll key to switch different dahal launcher to right- to launch， without having to open the spellbook to select the spell level and then launch

ATT2CHTranslationTeam:
This is just an optimization of the launch method, not a particularly big change.

ATT2CHTranslationTeam:
Because many players complain that they need to open the spellbook to select the level every time they cast a spell, and then use the dahal launcher to cast the spell, which is very torturous, especially in a multiplayer situation, the book will not pause when it is opened, and it is possible to be killed by monsters during this interval, so I optimized the way of casting a spell to avoid players having to open the spellbook multiple times.

ATT2CHTranslationTeam:
image
image
image

Piccomaster:
Ok thanks for the explanations, I will make a backup on my side and test that.

Piccomaster:
I'm trying to test with the fireball spell but it doesn't seem to work, even though I did the command /function att2:gameplay/dahal/launcher/initialize, and refresh the spellbook...
Edit : Ok I understand I need to remove the spell book, it's weird but convenient at the same time ^^

ATT2CHTranslationTeam:
> I'm trying to test with the fireball spell but it doesn't seem to work, even though I did the command /function att2:gameplay/dahal/launcher/initialize, and refresh the spellbook... Edit : Ok I understand I need to remove the spell book, it's weird but convenient at the same time ^^

Emm, hold on a sec. First, pull the latest PR, then give it a try. The previous one couldn't switch if you didn't select a spell level

PR #125, opened 2025-04-06
Fixed some bugs; boss kill detection is now stricter.

ATT2CHTranslationTeam:


Piccomaster:
Are you sure your boss detection system works? Because for example, there are exceptions, like for the elevator or the Golem, where the player doesn't directly kill an entity with their hands. Well, I suppose you've planned ahead, but you still need to be sure that everything works when the update comes out.

And I saw that there are inconsistencies in the bonus rewards depending on the time it takes the player to kill the boss, for example, for Kum... There are surely others.

Just know that you are in the process of thoroughly reviewing fundamental systems that have never been touched and which can therefore create major and very blocking bugs, so I trust you to test everything if I accept the PR.

And I'd like you to explain to me exactly what you changed in the rune system. Because I don't really understand why you even changed the main scoreboard?

In any case, once again, it's a lot of work, great job, but try to correct these small inconsistencies if there are any.
I'll have to put your name in a new credit category for the development of the map code haha ^^

ATT2CHTranslationTeam:
"I didn't modify the 'elevator' because it can't be checked using advancement. As for 'Golem,' I converted the 'instant damage' into a damage command for calculation, so there's no need to worry about that."

ATT2CHTranslationTeam:
Regarding the "reset" of the rune system:
Removing the scoreboard will directly clear all data under that score criterion, achieving the effect of a reset. This eliminates the need to manually reset each individual score entry, significantly reducing code volume.

Piccomaster:
Ok but now, in your latest commit, especially regarding combos, why did you remove this line?
scoreboard players remove @s KILLVALUE 1

It is an important line code, looks the go processing function of the leveling system :
execute unless entity @a[scores={KILLVALUE=1..}] run function att2:gameplay/leveling/monster/loot/actualizemaxclass
execute as @a[scores={KILLVALUE=1..}] run function att2:gameplay/leveling/monster/loot/sharedxp

ATT2CHTranslationTeam:
The experience gain mechanism has been optimized. In the previous system, the experience awarded to players was based on the data of the nearest monster to the player, rather than the actual monster they killed. This also applied when players killed bats—experience was still calculated using the nearest monster's data, which could lead to imbalances.

The new experience system features more precise detection, ensuring players can no longer gain excessive experience by killing bats.

ATT2CHTranslationTeam:
> The experience gain mechanism has been optimized. In the previous system, the experience awarded to players was based on the data of the nearest monster to the player, rather than the actual monster they killed. This also applied when players killed bats—experience was still calculated using the nearest monster's data, which could lead to imbalances.
> 
> The new experience system features more precise detection, ensuring players can no longer gain excessive experience by killing bats.

Maybe I'm overlooking something else, but "killvalue" doesn't do anything else in the system that detects experience gain from killing monsters

Piccomaster:
Yes indeed the system for detecting killed monsters was not perfect. Thanks for this correction but it doesn't change the problem that I said, it's that now the system remains functional permanently activated, since there is no longer any removal of the KILLVALUE value, right?

Look the go function processing "adventquest_functions\data\att2\functions\gameplay\leveling\monster\go.mcfunction" :
```
#############################################################################
#Made by Thundesrtruck														#
#use /funciton att2:gameplay/leveling/monster/go to initiate the processing	#
#############################################################################

execute as @e[tag=LVL0] run function att2:gameplay/leveling/monster/initialize/initnewmonster

execute unless entity @a[scores={KILLVALUE=1..}] run function att2:gameplay/leveling/monster/loot/actualizemaxclass

execute as @a[scores={KILLVALUE=1..}] run function att2:gameplay/leveling/monster/loot/sharedxp

execute as @e[tag=SUPER,team=hostile] at @s run function att2:gameplay/leveling/monster/super_elite_diff
execute as @e[tag=MEGA,team=hostile] at @s run function att2:gameplay/leveling/monster/mega_elite_diff
```

This line :
`execute as @a[scores={KILLVALUE=1..}] run function att2:gameplay/leveling/monster/loot/sharedxp`

If we not remove the value of KILLVALUE then the sharedxp system will be activated continuously 20 times per second...

ATT2CHTranslationTeam:
I'm not quite sure what else this "killvalue" is for, but it's useless for my optimized mob killing experience detection system, as'm using "advancement" to detect the mobs killed by the player, and experience is only awarded when the player kills a mob with the tag "class_scxxx", but I think I forgot to add the code to define "class_sc_xxx"... Anyway, I've tested this system in both single and multiplayer it works fine.

I forgot to modify this killvalue, it's not a big deal.

Piccomaster:
The KILLVALUE score was the central point of the system because it was what detected whether a player killed a mob. Look in the initialization:
`scoreboard objectives add KILLVALUE totalKillCount
`
In short, this was what triggered the sharing of experience. It's very important to comment it out in the "go" function if you no longer use this value because the player will continually have a KILLVALUE greater than 1 and will increase each time they kill an enemy.
Try to understand the system before making any major changes because it can affect other systems.
Yes it's not a big deal but even if what you changed works, other things may have been significantly impacted.

Piccomaster:
Since your last update on 
@ATT2CHTranslationTeam
 [Title Display Optimization XP File Repairs], the TITLE system doesn't work anymore... Why did you change that ?

ATT2CHTranslationTeam:
> Since your last update on @ATT2CHTranslationTeam [Title Display Optimization XP File Repairs], the TITLE system doesn't work anymore... Why did you change that ?

Mainly I found in the previous multi-person test, when there are two players, and the "consciousness" book is constantly refreshed, this "title will constantly switch between the "title" of the two players, so I made some modifications, but it seems that some other problems have occurred...

PR #126, opened 2025-04-08
fix spell system bug

ATT2CHTranslationTeam:


Piccomaster:
Why did you change the potion detection system, it was clearly efficient?

ATT2CHTranslationTeam:
> Why did you change the potion detection system, it was clearly efficient?

Because of the previous potion system, if the player drinks a potion with their off-hand, the system will not detect that the player drank a potion
And if the player switches items too fast while drinking a potion, it may cause the potion to not be recognized.

PR #128, opened 2025-04-11
fix horse seller bug

ATT2CHTranslationTeam:
I have to say, the spawn rate of mob invasions is really low... I currently started a new save, progressed the main story to Angband, and haven't had a single invasion spawn in 7 hours.


Piccomaster:
Ok I will increase the rate.

PR #129, opened 2025-04-12
fix recycle esc error

ATT2CHTranslationTeam:


ATT2CHTranslationTeam:
We could consider adding 'ULT' gear drops to the rune chests in the 'Mob Invasion' event

Piccomaster:
I thought about it too but the drop rate is too low and I don't want to do a balance according to an invasion level and the zone and in addition the mobs which appear in the invasion will already have a chance to drop the ult.

PR #130, opened 2025-04-13
fix minnor bug

ATT2CHTranslationTeam:


ATT2CHTranslationTeam:
To make it function properly, input is required.：
scoreboard objectives add LIMIT77 dummy
scoreboard objectives add SPELL34_CHEST dummy

PR #134, opened 2025-04-17
Optimized Boss Detection System

ATT2CHTranslationTeam:
The issue causing false "victory" triggers stems from the loading sequence when re-entering the game: players, chunks, and other entities load in a specific order, with entities typically loading last. This creates a brief window where entities are undetectable, mistakenly ending boss fights. Currently, this only affects bosses in "function att2:cinematic/misc/go". The solution involved converting these to command blocks placed near their respective boss arenas, which has been tested and resolves all reconnection-related issues.


ATT2CHTranslationTeam:
Due to the overly strict kill detection of "advancement" (e.g., wither or fire effects from certain spells, or spell26's iron golem killing the boss may not trigger progress detection, potentially causing issues), only some boss detection mechanisms have been modified.

Piccomaster:
The problem is that it won't work for some bosses.

We decided to use the command blocks in the center 777 77 777 precisely because we couldn't do otherwise due to the detection area being too large...

Let me explain: with Ouran, Serile, and other bosses, for example, the arena is far too large, and therefore, if the player only has chunk vision at 10, and the command block is no longer loaded, the system will no longer work at all.

You're reviewing complex systems that can cause major bugs and stuck totally the players. I can't accept this request, at least not until we're certain that every boss, every configuration in every part of the arena, is verified as safe and without major bugs.

ATT2CHTranslationTeam:
I have tested all these modifications with a new save, each part has been tested by quitting and re-entering, there is no problem, as for the of too far to load, I have set those parts to be “forceload”command, so there will be no problem.

ATT2CHTranslationTeam:
For example, in the serile boss battle, if the player exits and re-enters, it will cause the lightning ball to explode directly and skip the calculation the next lightning ball, which may lead to a series of problems such as getting stuck.


Or in the last serile boss battle place, after exiting and re-entering the door, it will cause unable to enter the boss door and other issues

Piccomaster:
Can you now forceload a command block? I didn't know that was possible; it was the major problem we had back then with arenas that were too big.

So can you confirm that even in the first arena of the Serile boss, where the player has to destroy the energy spheres, the detection still works?
And the same for the Ouran boss, everything works?

ATT2CHTranslationTeam:
Of course, I have tested all of these, and they all work properly.

"forceload add ~ ~" will force-load the current chunk.

Piccomaster:
Okay, fine, but that will increase the lag a little.

orian34:
Isn't it better to not use command blocks? They're pretty archaic and needing to forceload isn't too great.

Functions should handle things easily no?

ATT2CHTranslationTeam:
> Isn't it better to not use command blocks? They're pretty archaic and needing to forceload isn't too great.
> 
> Functions should handle things easily no?

Because there is a certain sequence of block and entity loading, if the area where the command block is located is too far away, it will cause an error that entities cannot be detected for a period of time after re-entering the game. As for the “tick” function file that comes with the data pack, I have also tested it, I don't know why some problems still occur.

Piccomaster:
Indeed, I think that even if a datapack function is active for chunks not forceloaded, it can cause problems due to the latency between the entities existing in the chunk and the chunk itself creating a lag between the game tick and the player.

Anyway, if the chunk is unloaded at the end of the boss, then it's fine. And if you confirm that boss fights with this system work well, then I'll merge everything.

We'll have to keep testing the map anyway. Ultimately, I don't plan on doing any updates yet; I'm adding exclusive content to complete this major update.

orian34:
> Because there is a certain sequence of block and entity loading, if the area where the command block is located is too far away, it will cause an error that entities cannot be detected for a period of time after re-entering the game. As for the “tick” function file that comes with the data pack, I have also tested it, I don't know why some problems still occur.

Mmh, I see yeah. Would it be possible to have the detection use a buffer when it's reinitialized through reloading?

PR #140, opened 2025-04-22
a system to change chest name/particle(C1-C10)+Reduce the triggering interval for mob invasions.

ATT2CHTranslationTeam:
This chest renaming system can modify the names and colors of chests c1 to c10. To assign different particle effects to c1-c10, simply adjust the particle colors in "misc\chesteffect\eye_effect\cXXXX_test".

if you modifying the particle effects for c1-c10, please update the chest particle effect commands in "misc/go".

To enable particle effects for chests c1-c10, the chest particle effect code in "misc/go" should only contain:
execute as @a at @s run function att2:gameplay/misc/chesteffect/eye_effect/go

Regarding the trigger interval for mob invasions,, our testers barely encountered this event at all. Its presence is practically negligible, so i have no choice but to reduce the trigger interval.


orian34:
Oh that's great yeah!

ATT2CHTranslationTeam:
"I'd also like to mention that some early-game chests, especially those in towns like Ryliaeth and Meleim, have really poor loot. There's also a high chance of getting nothing when opening them. I think it would be better to increase the drop rate a bit. The current probability is quite discouraging for players who enjoy collecting items—at least many people I know have complained about why so many chests are empty..."

"For example, change the loot roll values for C1T1 from:
"rolls":{"min":0,"max":1},"bonus_rolls":{"min":0,"max":0}"
to:
"rolls":{"min":1,"max":2},"bonus_rolls":{"min":0,"max":0}"
like this."

ATT2CHTranslationTeam:
> This chest renaming system can modify the names and colors of chests c1 to c10. To assign different particle effects to c1-c10, simply adjust the particle colors in "misc\chesteffect\eye_effect\cXXXX_test".


To be honest, modifying chest names in the map should ideally be done using a map editor, but after some research, I couldn't figure it out. So, I had no choice but to use a data pack to detect and modify the chests in front of the player in real-time instead.

orian34:
Yeah, it can be disorienting to have empty chests with shimmer. Maybe ensuring at least a baseline of some coins/basic misc item?

Piccomaster:
Hmm, okay, why not, It's a good idea. Anyway, there are too many chests to change them manually, so as long as it doesn't cause too much lag, I'm fine with that.

Regarding the invasions, I haven't tested them yet, but if you have people who's already played the latest version of the map, I trust you.

As for the C1T1 chests, I'll take care of that, but originally it was perfectly normal to find nothing in the chests, so if I change it, I'll increase almost nothing, at most:
"rolls":{"min":1,"max":1},"bonus_rolls":{"min":0,"max":1}"

Piccomaster:
Your latest addition to spells, prevents me from accepting this PR.

I understand that support spells like the health regeneration "healing" can be an aura that heals allies, but that's really not their purpose for other spells...

It's just too powerful if it gives everyone the same effect. That's not how spells were designed.

If this change is necessary, then we need to think about some kind of synergy that unlocks by completing a quest, or else just the last level of the spell would allow this. But giving players the ability to buff everyone directly I don't want that sorry.

ATT2CHTranslationTeam:
My idea focuses on adding multiplayer spell interactions, as current spells mostly only affect the caster, resulting in limited utility in cooperative play. This makes many spells feel ineffective in multiplayer scenarios.

For example, while one player engages in frontline combat, another could provide healing or strength buffs from the rear. Adding such interactions would make battles more dynamic and engaging.

Of course, its effects on others could be appropriately nerfed - having identical potency would likely be overpowered.

Piccomaster:
Yes, or perhaps mitigating the effect on other players would be better. But I want a counterbalance to this change because it will significantly reduce the difficulty for players in multi. I'll think about that.

ATT2CHTranslationTeam:
I've rebalanced the spell effects for multiplayer—their potency is now approximately 70% of the self-cast version.

PR #142, opened 2025-04-29
new weapon system

ATT2CHTranslationTeam:
To make it work properly, you need to input:
"function att2:gameplay/equipment/initialize"
and
"function att2:gameplay/legendary/initialize"

The damage command utilizes macro functions, which can transfer numbers stored in storage to the damage command, enabling scoreboard-based damage value modification without requiring individual commands for each damage instance.

The axe's armor enchantment reduction works as follows: In reality, when the total Protection enchantment level on armor reaches 20 or higher, it becomes completely ineffective. Therefore, when the axe attacks an enemy, I first readjust the Protection enchantment before reducing its level (each Protection level provides 4% damage reduction), without decreasing the enchantment level excessively. As for the armor value itself, armor's damage reduction is more noticeable against low damage but becomes increasingly ineffective against high damage, so I simply reduced it by 50%.


ATT2CHTranslationTeam:
That means we can use macro functions to optimize the spell damage files, making them less bloated and easier to modify


Additionally, we can use attribute @s... to more flexibly modify players' max health (allowing 0.5 or 1-point increments), eliminating the need for effect give @s health_boost (which only provides 4 HP/two hearts).

Piccomaster:
Nice work, but I'd need more details on what the bonuses do for each weapon type you added, for the changelog.
I'll test that later.

ATT2CHTranslationTeam:
For nearly all weapons, I base the damage on the player's current potential damage output. 
Taking the sword as an example: 
if the player currently has a strength buff level of 10 and is wielding a weapon with a base attack damage of 20, then:
The first counterattack damage after blocking would be (10×3 + 20)/2 = 25 damage
The fifth counterattack damage would be (10×3 + 20)×1.5 = 75 damage (same as the player's leaping strike damage)

Axe and hammer don't have extra damage multipliers, but the hammer's AoE damage equals the player's total current attack damage.

The dagger's third attack deals damage equal to (Player's Current Total Attack Damage) × [1.5 + 0.04 × (Player's Current Haste Level)].(Since this system's damage will take the maximum value compared to the player's regular attack damage, this strike must exceed the total attack damage to ensure it actually deals damage.

The spear's area damage equals (Player's Total Attack Damage)/4 × [Speed Buff Level × 0.2 + 1].)

ATT2CHTranslationTeam:
While the functionality is implemented, I can only do my best to make the weapon attack effects look more visually appealing... To be honest, I'm not very skilled with special effects. If you have better ideas, further optimizations could be made—especially for the axe effect. I can't think of a good way to add an effect that shows when a monster's armor is already broken...

orian34:
> While the functionality is implemented, I can only do my best to make the weapon attack effects look more visually appealing... To be honest, I'm not very skilled with special effects. If you have better ideas, further optimizations could be made—especially for the axe effect. I can't think of a good way to add an effect that shows when a monster's armor is already broken...

What if you used a similar effect as the quest mark on NPC? But instead it's a broken armor icon

Piccomaster:
Great, thanks. I did test it, and it's missing a bit of visual flair. I was thinking the same thing as Orian, but it would still be better to emit particles around the mob; it's just that on super elites, it will be less noticeable.

In any case, it does buff the player who plays melee a bit. The spells will be a little less attractive for players, but given the recent spell buffs, it should be balanced.

However, I'm still a little worried that this version of the game will be made much easier than before. I'll try a new adventure on the map (maybe not the whole thing), but I think it will be essential.

Piccomaster:
I tried to make a summary for the changelog of the additions of the new weapon system and I would like to know if it is correct (Because looking at the code, it seems to me that some things have changed a bit compared to the original premise):

- Added an effect for each type of weapon the player is wielding. These effects are dynamically calculated based on certain active buffs and the number of hits landed on the opponent.
- Sword: A counterattack after blocking, dealing more damage based on the strength and the number of consecutive hits landed on the opponent.
- Hammer: A circular area effect on each hit, dealing damage equal to the player's current damage total.
- Dagger: A third, more powerful hit with a multiplier based on the player's attack speed.
- Spear: A cone-shaped area effect with a multiplier based on the player's movement speed.
- Axe: An effect that reduces opponents' armor.

Is this a good summary?

ATT2CHTranslationTeam:
> I tried to make a summary for the changelog of the additions of the new weapon system and I would like to know if it is correct (Because looking at the code, it seems to me that some things have changed a bit compared to the original premise):
> 
> * Added an effect for each type of weapon the player is wielding. These effects are dynamically calculated based on certain active buffs and the number of hits landed on the opponent.
> * Sword: A counterattack after blocking, dealing more damage based on the strength and the number of consecutive hits landed on the opponent.
> * Hammer: A circular area effect on each hit, dealing damage equal to the player's current damage total.
> * Dagger: A third, more powerful hit with a multiplier based on the player's attack speed.
> * Spear: A cone-shaped area effect with a multiplier based on the player's movement speed.
> * Axe: An effect that reduces opponents' armor.
> 
> Is this a good summary?

The weapon effects are implemented correctly, but the hammer requires a jumping attack to trigger, while the spear's area damage and knockback effects only activate when attacking during sprinting or within 0.5 seconds after stopping a sprint

Piccomaster:
Oh, indeed, I forgot a few details. Thanks.

PR #145, opened 2025-05-04
Spell Damage System Update

ATT2CHTranslationTeam:
Use macro functions to recalibrate the spell damage calculation method
before use there are some function to input:
scoreboard objectives add SPELL10_OWNER dummy
scoreboard objectives add SPELL5_OWNER dummy
scoreboard objectives add SPELL6_OWNER dummy
scoreboard objectives add SPELL7_OWNER dummy
scoreboard objectives add SPDG dummy

To modify spell damage for levels 1-10, simply adjust these two values in the 'damage_cal' file:
scoreboard players set minDG SPDG XXX
scoreboard players set maxDG SPDG XXX

These two values define the spell damage range. After input, the damage per spell level will be distributed according to the calculation formula at the top of the file, then passed to the damage command via macro functions.


Piccomaster:
Nice work indeed, but it might require some balancing, as the early levels are way too powerful. Especially for Earthquake, for example, which goes from 5-10 damage per wave to 50 haha...

I'll take a look at that and adjust it based on what we decided last time for the damage adjustment.

PR #147, opened 2025-05-06
fix exploit book

ATT2CHTranslationTeam:
The operating procedure remains the same as before


ATT2CHTranslationTeam:
only once
function att2:gameplay/misc/exploit_book/initialize
function att2:gameplay/speceffect/random/initialize
/////
function att2:gameplay/misc/exploit_book/add_test
function att2:gameplay/misc/exploit_book/go with storage att2:exploit_book

"/function att2:gameplay/misc/exploit_book/give_book" can properly grant you the exploit_book

ATT2CHTranslationTeam:
Now each level up increases maximum HP by 1.2 points.
All other values remain unchanged from the original

ATT2CHTranslationTeam:
Updated the random number selection and calculation to use these two integral terms. Going forward, please use these two integral terms for all value selections and calculations, as creating separate integral terms for each function is too cumbersome

Piccomaster:
Ok, great initiative.

PR #149, opened 2025-05-08
mimic system

ATT2CHTranslationTeam:
The system is functional, but many details remain unoptimized—like the 'mimic' loot and interaction feedback mechanisms. I'm currently stuck for ideas on these.

Mechanics of the Mimic System:
When opening a chest, there is a 0.05% chance to spawn a mimic. The mimic's strength scales with the quality of the opened chest. The mimic does not attack players. When players attack the mimic, it triggers one random event out of 10 or more possibilities - these could be minor traps (generally non-lethal) or have a chance to trigger beneficial events (such as currency or XP rewards).

The mimic's quality determines the percentage of HP lost per player attack. Higher quality allows more attacks before death and grants greater rewards upon defeat.

Before use, need first enter: "
function att2:gameplay/enveffect/mimic/initialize
function att2:gameplay/speceffect/random/initialize
#This command gives you a chance to trigger summon a mimic.
function att2:advancement/test_all/treasure/lootchest_c10/tier10


Piccomaster:
Oh yeah, what a system!
That's a lot of functions for such a minor feature. But as we say back home, it's in the details that all the difference is made, good job.
I'll test this in my entire walkthrough of the map.

PR #168, opened 2025-05-24
Optimize spell cooldown calculation

ATT2CHTranslationTeam:
"Cooldownrush" is my new enchantment-related content, so let's just assume it doesn't exist for now

### DIFF
--- a/adventquest_functions/data/att2/functions/dialogs/gameplay/dahal/remain/spell1.mcfunction
+++ b/adventquest_functions/data/att2/functions/dialogs/gameplay/dahal/remain/spell1.mcfunction
@@ -7,6 +7,8 @@ scoreboard players operation @s remain_time = @s COOLDOWN1
 scoreboard players operation total remain_time = 20 remain_time
 scoreboard players operation total remain_time += BonusCooldown RUNE
 scoreboard players operation total remain_time += BonusCooldown RUNE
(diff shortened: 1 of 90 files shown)


Piccomaster:
This is for a version higher than 1.20.2, so it won't appear in the 5.0.0 update? If so, I won't accept the PR right away; Because I don't want to mix everything up, I don't want any changes related to things that won't appear in the update 5.0.0.

Once the planned update is released, I'll create another "upgrade 1.21" branch to avoid uploading new changes that will obsolete Minecraft 1.20.2 commands to the current "develop" branch.

ATT2CHTranslationTeam:
Oh, no need to worry about this part—it just adds two extra scoreboard operations with zero impact. I'm simply too lazy to remove them (^^

Piccomaster:
Yes, I've seen it, I understand but I don't want any additions to the current "develop" branch regarding version changes. I want everything related to versions after 1.20.2 to be in a separate branch.

I don't know how long the upgrade to 1.21 will take and which systems will actually be impacted. I don't want to mix anything between versions; I prefer to have clean versioning and a well-organized Git repository to better see the version history if we need to rollback or not.

ATT2CHTranslationTeam:
OK, you're right—it's just a minor optimization anyway. Let's figure it out later then

PR #193, opened 2025-06-27
Updated pack.mcmeta

ElGrenier:


ElGrenier:
Nothing too fancy, just updated pack.mcmeta so i don't have any problem in VsCode nor ingame

PR #197, opened 2025-06-28
updated some function

ElGrenier:
Updated some function, removed some unused function (thoses to sell Dahal weapons), with this push, the whole datapack has gotten under the 14k bug ! (arround 1000 bug resolved by this bugfix, litterally)


ATT2CHTranslationTeam:
Oh no, a very bad new—you didn't completely remove the 's' from all datapack structure names, which has now caused duplication issues...
like this...
DFYQ`UYI ~2JGNKG)@32X{X
Seems a bit complicated

Piccomaster:
I reverted the last PR of Grenier.

Piccomaster:
@ElGrenier Je suis revenu en arrière en annulant tes deux dernières Pull Request car ça a fait doublon de toutes les fonctions du jeu (voir le message au-dessus pour plus de précisions).

Piccomaster:
It's fixed.
But I would like to understand @ATT2CHTranslationTeam , what exactly was the purpose of removing the 's' from all datapack structure names ?

ATT2CHTranslationTeam:
Because the new version's datapack structure is just like this... Mojang changed a lot of naming conventions...
Otherwise, the datapack won't be recognized correctly.

orian34:
> Because the new version's datapack structure is just like this... Mojang changed a lot of naming conventions... Otherwise, the datapack won't be recognized correctly.

Indeed that is the case, vanilla changed a lot of folder paths that way.

Piccomaster:
Oh ok Mojang...

PR #206, opened 2025-07-03
fix some bug

ATT2CHTranslationTeam:
Some teleport commands cause NPCs to stand up instead of staying seated, so they need to be manually reset to a sitting position.


ATT2CHTranslationTeam:
OK, I've finished testing all main quests. The required updates are listed in the 'update' file—run these two commands via looping command blocks in-game. Note: Due to chunk loading issues, you'll need to physically traverse all relevant areas to trigger updates

Piccomaster:
Ok great, anyway it was planned to completely go through all the chunks of the map to update them by loading them into the latest version of Minecraft.

Piccomaster:
But to do everything in one go, wouldn't it be better to also prepare a function for loading command blocks for boss arenas that weren't working because of chunks? If I remember correctly, you created commands located in the folder:
"att2:function\cinematic\misc\reset_cb"

And that way, during the players' adventure, these commands will no longer need to be activated and can be deleted?

ATT2CHTranslationTeam:
Yes, I'll have this ready, 1.21.5 needs to update some other command blocks, and I'll get them in order when the time comes

Piccomaster:
Ok I'll wait until you've finished that before doing the full map loading in 1.21.5.

PR #209, opened 2025-07-05
fix  bug | Optimized Mob Spawning System

ATT2CHTranslationTeam:


ATT2CHTranslationTeam:
SPAWNER_LIMIT is a mechanic I added for small chairs.：When a player sits on a chair, no monsters will spawn within a 30-block radius of them. This ensures that players won't be ambushed by mobs while fishing

Piccomaster:
Lots of questions before accept this PR:
- Why did you add the impossible Minecraft advancements files again for Disable All Crafting Recipes & Vanilla Advancements? I had modified a rule to specifically disregard this, no need of all theses files anymore...
Look at the pack.meta file for advancements:
```
{
  "pack": {
    "pack_format": 18,
    "description": "Across The Time 2 advancements by Adventquest for Minecraft 1.20"
  },
  "filter": {
    "block": [
      {
        "namespace": "minecraft",
        "path": "advancements/.*"
      }
    ]
  }
}
```

- How does this chair system work? Aren't you afraid that this system could be exploited to bypass many combat passages?

- Regarding the addition of new advancement with achievement point gains: be sure to update and calculate the total number found on line 14 in the file: att2/functions/advancement
/initialize.mcfunction.
You also need to change this total, which is found here on line 97:
att2/functions/cinematic/sidequest
/go.mcfunction

- One last remark, as usual, congratulations given the huge changes made, but above all, don't forget to constantly update the CHANGELOG, as I did because you have to be organized and list the changes well, such as the total number of Exploits points to earn compared to the previous version. In short, look at the old changelogs for inspiration.

ATT2CHTranslationTeam:
Oh, I didn't notice this advancement detail earlier—my bad.

ATT2CHTranslationTeam:
Small chair system: 
Right-click while holding the small chair to place it (essentially an invisible minecart). 
Players can then "sit" on the "chair". 
It won't cause any sequence breaking because it can only be placed beneath the player.
 I've also added a mechanic: 
when a player is sitting on the chair, their entity interaction range becomes 0, meaning players can't place multiple chairs to achieve "fast movement".

PR #214, opened 2025-07-09
update advancement | game Percentage

ATT2CHTranslationTeam:


Piccomaster:
Oh crap, just now 2 minutes ago I fixed the "scale"... So the PR is in conflict.
I think you can remove the commit "Fixed the syntax error for attribute scaling."

Piccomaster:
Also, are you aware that you've removed the entire completion system that calculates the percentage of whether the player completes the game 100% or not?

Furthermore, there aren't exactly 300 MAINQUEST points unlocked by the player, because sometimes there are jumps when the player moves from one important moment to another in the game's story.
For example, when the player, after depositing the third Space Gem, goes directly from MAINQUEST score 42 to 51, there are no MAINQUEST score 43, 44, 45, 46, 47, 48, 49...

Piccomaster:
One last point, I think we should also stop adding empty and unnecessary lines in the code at the end of each function, because it not only makes the project heavier but also in the commits we don't see any real changes made haha.

ATT2CHTranslationTeam:
> Also, are you aware that you've removed the entire completion system that calculates the percentage of whether the player completes the game 100% or not?
> 
> Furthermore, there aren't exactly 300 MAINQUEST points unlocked by the player, because sometimes there are jumps when the player moves from one important moment to another in the game's story. For example, when the player, after depositing the third Space Gem, goes directly from MAINQUEST score 42 to 51, there are no MAINQUEST score 43, 44, 45, 

I know, but it's just a system for calculating percentages. Even if the mainline numbers don't completely cover the range of 1-300, after completing all the mainlines, it will still become 300. This means that such modifications will only make the achievement trigger less timely, and the final result will be the same

> Also, are you aware that you've removed the entire completion system that calculates the percentage of whether the player completes the game 100% or not?
> 
> Furthermore, there aren't exactly 300 MAINQUEST points unlocked by the player, because sometimes there are jumps when the player moves from one important moment to another in the game's story. For example, when the player, after depositing the third Space Gem, goes directly from MAINQUEST score 42 to 51, there are no MAINQUEST score 43, 44, 45, 46, 47, 48, 49...

I know, but it's just a system for calculating percentages. Even if the mainline numbers don't completely cover the range of 1-300, after completing all the mainlines, it will still become 300. This means that such modifications will only make the achievement trigger less timely, and the final result will be the same

And the previous calculation method's mainline progress percentage was also derived from the mainline number.. So there's actually no difference

PR #215, opened 2025-07-09
update

ATT2CHTranslationTeam:
OK, I'll just remove all the useless line breaks at the end of files first... Only the useless line breaks are removed, nothing else.


Piccomaster:
It's strange, I have the impression that some changes have already been made in the past with certain commits... Well, I hope this time there won't be any more commits affecting all the functions features just for a blank line, because that also prevents other people from contributing things to the project without creating conflicts.

In any case, I'm looking forward to testing the new enchantment and fishing system. The balancing phase will take some time so that the difficulty isn't affected too much by making the game too easy.

PR #216, opened 2025-07-10
mainplayer name modification | add fishing system loottable

ATT2CHTranslationTeam:
Players can now manually change the "mainplayer name," supporting more characters and languages.


ATT2CHTranslationTeam:
So far, I’ve only redesigned a few of them. Honestly, some Mythical gear—especially Blindshield and Ultima—aren’t very well-designed. They lack distinctive features or engaging mechanics. Right now, my creativity is running dry. Just reworking these few pieces has nearly drained me.

The design process requires:
Avoiding overlap with the enchantment system,
Implementing interesting mechanics,
But not cramming in too many mechanics at once.

This whole thing is way harder than I expected...

ATT2CHTranslationTeam:
Respawn point sharing causes weird issues and ruins the multiplayer experience (e.g. when one player triggers a respawn point in Billgart , while another player is farming bosses in Angband - if the Angband player dies accidentally, they'll respawn in Billgart's dimension, which feels extremely unnatural). 
Main/side quest respawn points now all have range checks implemented, so there's no need to worry about players progressing main quests spawning in wrong locations.

Piccomaster:
Regarding the legendary weapons — yes, I expected that balancing them would be tricky, especially with your enchantment system. It’s important that your system doesn’t allow them to be affected.
We absolutely must avoid breaking the game balance. Mythical weapons are not meant to ruin the game’s difficulty by making all combat encounters trivial.

By the way, I hope you haven’t changed their core mechanics, because I noticed you modified quite a few files, like Rukyrion's and others.
Sure, I’m okay with improving them — but let’s not go too far and distort the original effects that Thundesrtruck and I carefully designed.

As for the checkpoints, yes, the map was originally built as a co-op experience, not a full-blown MMORPG, so players were meant to stay relatively close to each other from one quest or dimension to another.
That said, the changes you’ve made in that area are definitely appreciated.

Same goes for the player name change — why not, even if it clearly involved a massive amount of work.

On the other hand, I don't understand why you removed the "LVL" from the stat scores, for example, what's the point of changing "SPD_LVL_EXT" to "SPD_EXT"? You're adding unnecessary work, right?

I’m really impressed by the amount of work you’ve done — you’ve brought in so many new things. One piece of advice: if you ever feel stuck or uninspired, take a break. Step back and reflect before coding a new system or feature you want to implement into the game.
Especially since I can only see the changes after the fact — and even then, I haven’t tested any of the new gameplay features or additions yet. I honestly don’t know what kind of impact they could have on the map — I can only give you my opinion based on impressions from the code, as a kind of supervisor. So I can’t really give you any proper feedback on the actual in-game changes.

So take your time — there's no need to rush this update. It’s precisely because I wanted version 5.0.0 to be solid and well-polished that we can now afford to take our time.

ATT2CHTranslationTeam:
"SPD_LVL_EXT" was modified because I made some changes to the attribute point scoring system during attribute system optimization, so these elements needed updating..

Regarding the modifications to Legendary Equipment:

I’ll definitely be careful with the changes. The main reason for optimizing this system is that I’ve noticed many players don’t have a very positive impression of Legendary Equipment—my own experience with the system hasn’t been great either. While it might have felt fresh back in 2020, by today’s standards, it’s fallen quite far behind.

Piccomaster:
But for 5.0.0, you had already upgraded the mythical items, and from then on, these items had become quite powerful. Why make them even more powerful?

ATT2CHTranslationTeam:
> But for 5.0.0, you had already upgraded the mythical items, and from then on, these items had become quite powerful. Why make them even more powerful?

That "upgrade" was just a minor optimization I made long ago - it never actually addressed the core issues with mythical gear. The current mythical equipment effects are far too weak, often less appealing than regular equipment stats.
The main problems are:
They lack interesting mechanics
They don't provide meaningful stat improvements to enhance gameplay
What I want to solve is making players actually use these mythical items after obtaining them, rather than treating them as mere decorations or collectibles. When players work hard to complete various side quests for a piece of gear, it should feel powerful - at least not in its current underwhelming state...

PR #218, opened 2025-07-12
Bug fix + removed useless files

ElGrenier:


ElGrenier:
Only 309 bugs + 86 Warnings ! 

What need update still is : 
summon_signs mcfunction to be updated
seal effect to be updated
Give of Legendary weapon to be updated

ATT2CHTranslationTeam:
Could the ']' in "mainplayer/summon's id" have been accidentally modified? It seems that part should be correct as is.

PR #219, opened 2025-07-13
update kinuil/sastr

ATT2CHTranslationTeam:


Piccomaster:
That makes a lot of functions for mythical weapon effects ^^. In any case, I'm really looking forward to testing these improved effects.

PR #221, opened 2025-07-15
Dialogue repair and guidance addition

ATT2CHTranslationTeam:
Dialogue repair and guidance addition


Piccomaster:
I have the impression that this PR doesn't take into account the latest fixes made by Grenier in his last PR, which I accepted. We've noticed the addition of unnecessary files that had been deleted, such as the spectral weapon items in the "showcase" folder for sales...

Are you sure you've pulled the latest fixes before to mak this changement request?

Piccomaster:
Ah, okay, I looked at each commit, in the right order. It's good after all.

PR #223, opened 2025-07-15
fix loottable/ fix minor error

ATT2CHTranslationTeam:


Piccomaster:
Why did you divide the cooldown time limit for the fireball level by 5? It's just really too much and allows the level to evolve too quickly, it unbalances the following levels too much...

ATT2CHTranslationTeam:
SPELL1_TIMER is not the cooldown for fireballs, that is a scoreboard objective used by a system to prevent knockback on players when fireballs explode.

Piccomaster:
Oh I guess I wasn't fully awake lol.

PR #240, opened 2025-07-20
fix cooked recipe | update all sidequest quest item display

ATT2CHTranslationTeam:


ATT2CHTranslationTeam:
Only updated main/side quest items in 'function\gameplay\intake'.

PR #241, opened 2025-07-21
fix advancement nbt->component| add attribute display system

ATT2CHTranslationTeam:
Currently, elite mob kill detection is all handled in "att2_test:triggerkill"


ATT2CHTranslationTeam:
like this
image
and
image
The boss bar's progress represents the player's current Dahal value.

And players can choose to display any attribute they want.

Piccomaster:
> The boss bar's progress represents the player's current Dahal value.
> 
> And players can choose to display any attribute they want.

Great idea, It was something we wanted to do but we were afraid that with the boss bar and enemy bar displays it would be too cluttered.

ATT2CHTranslationTeam:
> Great idea, It was something we wanted to do but we were afraid that with the boss bar and enemy bar displays it would be too cluttered.

We could actually consider creating a completely new monster health display system. For non-boss monsters, health bars could appear above their heads, allowing players to see each monster's health bar individually. The current system only shows the nearest monster within visual range, which sometimes doesn't select the one players want to track, and it can only display one monster at a time.

orian34:
> > Great idea, It was something we wanted to do but we were afraid that with the boss bar and enemy bar displays it would be too cluttered.
> 
> We could actually consider creating a completely new monster health display system. For non-boss monsters, health bars could appear above their heads, allowing players to see each monster's health bar individually. The current system only shows the nearest monster within visual range, which sometimes doesn't select the one players want to track, and it can only display one monster at a time.

Isn't it possible to have multiple boss health bars? So increasing it to having 3 or 5 at once could solve it no?

ATT2CHTranslationTeam:
I think that would be very bad, because if you use bossbar to display the health bars of one or more ordinary monsters, players can't really distinguish which bossbar corresponds to which monster, especially when there are many monsters..

Piccomaster:
If you think it's feasible, why not. At least, with the version we started with, we thought about a lot of systems, but they were far from optimal and most of them were impossible.

I'd also like to take this opportunity to suggest putting more information above enemies' heads, because the current symbols, which only indicate the enemy's difficulty, are not informative enough.

Perhaps we could add an option allowing players to display whatever they want, for example:
- The enemy's class number,
- Whether it's an Elite, Super Elite, or Mega Elite (keep the color green, orange, or red),
- And if there's no other way, display the HP (but I don't think that's feasible dynamically).

PR #248, opened 2025-07-23
fix enchantments predicate error

ATT2CHTranslationTeam:
I'm not entirely sure - my eternan equipment section isn't throwing any errors. Did you update the adventquest_item_modifier file during your testing?


Piccomaster:
Indeed, I even tested it in a bad version hmm sry...

PR #250, opened 2025-07-25
update healthbar system|fix some advancements (kill trigger)

ATT2CHTranslationTeam:
image
I'm a bit unclear... I'm really confused right now, umm. In the current monster level system, there are two "levels": one is the monster's own CLASSLEVEL, and the other is LEVELDIFF, which adjusts based on the difference between the player's level and the monster's level. I also wanted to try preserving symbols like ✬, but I'm not sure how to display them... Because right now, to reduce performance overhead, I can only use two display entities—the item display entity is already being used for the health bar, so I can't add any other textures to it, and the text display entity's content is already quite packed..


ATT2CHTranslationTeam:
# Health Display System  

## Player Settings  
- Adjustable display range for health bars  
- Toggle options for:  
  - Numeric HP values  
  - Percentage-based HP  
  - Monster level indicators  

## Color Coding  
### Regular Monsters (by ClassLevel)  
`Green` | `DarkGreen` | `Yellow` | `Gold`  
`Red` | `DarkRed`  

### Elite Monsters  
- **Elites**: `Blue`  
- **Super Elites**: `DarkBlue`  
- **Mega Elites**: `DarkAqua`  

### Detection Mechanism
- The system checks twice per second whether the attached enemies of the item display entity exist.

- **Combat Handling**:  
  Every time a player attacks or kills a monster:  
  1. Processes the HP of the damaged monster.  
  2. Forces the health bar to glow for **0.5 seconds** after the monster takes damage, providing visual feedback to the player.

Piccomaster:
Well, for me it's perfect like that, in fact we shouldn't display too many entities and lose performance.

In fact the LEVELDIFF just allows for the player to see the difference with a custom name color on entites and so to display the difference in power of the mobs which have a CLASSLEVEL (the level of the enemy), with the GAMELEVEL (the real level of the player).

PR #251, opened 2025-07-25
add Automatic Pathfinding System[Beta] |fix healthbar intialize

ATT2CHTranslationTeam:


ATT2CHTranslationTeam:
# Automatic Pathfinding System  

**Principle**:  
A large number of road markers (`road_marker`) are placed on the map, and a specific destination is set. After selecting the destination, a `<route entity>` is generated, which teleports to the nearest road marker.  

Logically, this road marker is isolated from the path network, and the system evaluates the "score" of other connected road markers to determine whether they can lead to the destination. If a valid path is found, the route entity teleports to the next road marker. Eventually, a complete route is generated.  

*Note*: The "score" used to determine path validity is manually configured, which may lead to minor errors. Currently, this is a work in progress—only basic testing has been done, and most points are functional. The system is not yet complete, as it still requires integration with a **Sylberland map display system**, which may take some time.  

---  

### **Current System Setup**  
1. **Initialization**:  
   - Place a command block loop near coordinates `777 77 777` and run:  
     ```mcfunction
     function att2:gameplay/misc/map/system_set/summon_marker/summon
     ```  
   - Then, execute:  
     ```mcfunction
     function att2:gameplay/misc/map/initialize
     ```  

2. **Route Selection**:  
   - Choose a destination by running:  
     ```mcfunction
     function att2:gameplay/misc/map/route_select/(endpoint)
     ```  
   - This will also spawn a vehicle:  
     ```mcfunction
     function att2:summon/ride/(vehicle)
     ```  

3. **Starting Navigation**:  
   - Ride the vehicle to the starting point.  
   - Right-click to sit on the invisible minecart above the vehicle, and the automatic pathfinding will begin.

Piccomaster:
It's impressive, I read a little of the code and in the part of the "score_compare" folder, it seems that there is a lot of redundancy  lines of code, if you ever finalize I think that this part can be optimized.

ATT2CHTranslationTeam:
I think that part might be difficult to optimize further. Most of it involves scoreboard operations, which don't cause significant lag. Besides, those scores control pathfinding directions and require manual configuration for each case...
At least for now, I don't want to make further changes to them... Writing pathfinding scores was truly painful, especially for looped routes where I had to manually calculate optimal paths

Piccomaster:
However I have questions about the initialization, are you sure you have to run the following commands in a loop?
`#forceload
function att2:gameplay/misc/map/system_set/forceload/run
#summon
function att2:gameplay/misc/map/system_set/summon_marker/road_point`

This will summon entities in loops and force the loading of a good portion of the map chunks permanently. It will just make the map crash...

ATT2CHTranslationTeam:
This is only the current testing implementation. All generated entities are UUID-tagged and won't respawn. As for chunk loading, I'll eventually integrate the force-load process into the path calculation function - no persistent force-loading will be needed during normal operation.

Piccomaster:
> Writing pathfinding scores was truly painful, especially for looped routes where I had to manually calculate optimal paths

Yes, that's why I thought you must have gone to a lot of trouble doing this.

But I still don't understand one thing, because all files like these have the same code inside:
function/data/att2/function/gameplay/misc/map/route_cal/score_compare/100/101.mcfunction
function/data/att2/function/gameplay/misc/map/route_cal/score_compare/101/102.mcfunction
function/data/att2/function/gameplay/misc/map/route_cal/score_compare/102/103.mcfunction
function/data/att2/function/gameplay/misc/map/route_cal/score_compare/103/104.mcfunction
Etc...

And these files also have exactly the same lines of code :
function/data/att2/function/gameplay/misc/map/route_cal/score_compare/100/98.mcfunction
function/data/att2/function/gameplay/misc/map/route_cal/score_compare/101/100.mcfunction
function/data/att2/function/gameplay/misc/map/route_cal/score_compare/102/101.mcfunction
Etc...

ATT2CHTranslationTeam:
I'm not entirely sure how to explain this... It's quite abstract. When the system runs, it detects armor stand UUIDs along the path, then uses the last digit of the UUID to select the corresponding detection file. You can think of them as waypoints on the same route - that's why they share identical destination scores. These components are all essential. During auto-pathing, the system must check nearby waypoints in score order to adjust momentum direction... and all these scores are stored on the waypoint entities themselves

Piccomaster:
Ok I'm starting to understand, but if it refers to the same scores it might as well just have a mcfunctions file that refers to this one with all the same scores. 
Afterwards I have trouble realizing the entire structure and the functioning but ok I'll let you finalize all that, I'll test when it's completely finished. Especially since I haven't gone over the whole map yet, I prefer to wait until the major projects are finished to be sure that no other command blocks on the map will have to be changed.

ATT2CHTranslationTeam:
image
The basic execution works like this: When you stand at point X and select destination Y, the pathfinding entity starts from X, checks linked entities' scores to verify if they lead to Y, thus forming a route. You'll notice all entities along the path share the same score value because both ends connect to the same route.

Piccomaster:
Ingenious! But it makes for a very heavy system, it must have taken you a lot of time to make it.

PR #260, opened 2025-07-31
update sq kill trigger |fix map unlock dialogs|fix runeword translate key

ATT2CHTranslationTeam:
## Visual Customization Options

**New Settings Dialog**:  
`function att2:dialogs/gameplay/misc/eye_effect/select`

**Adjustable Visual Elements**:
- Health bar system
- Item particle effects  
- Attribute displays
- Dynamic lighting
- Hero title particle effects


ATT2CHTranslationTeam:
mobs_invasion system:
Previously, it was mainly due to the fact that no matter where the player was, this system would make a selection, resulting in "idling" most of the time, so the frequency of spawning monster groups was particularly low

ATT2CHTranslationTeam:
I'm currently conducting full game flow testing, which will result in numerous minor content adjustments

Piccomaster:
Ok very well, it is clear that the map needs test, but I cannot for the moment provide the map having loaded all the new chunks because I do not really know if it is all finished at the level of the "update" folder.

ATT2CHTranslationTeam:
My map testing baseline uses version 5.0.1 as a foundation, with all chunks updated via the 'update' file. As for whether this 'update' file is complete... I suggest waiting until my initial testing concludes before you doing a full chunk update, since I'm unsure if further adjustments are needed. Therefore, this 'update' file will remain in a perpetual state of modification - at least for the near future, it won't be a finalized version.

Piccomaster:
Ok all right,, thanks for the reply, I'll wait.

PR #269, opened 2025-08-12
Maze Mechanic Optimization

ATT2CHTranslationTeam:
### New Feature: Maze Torch
- **Acquisition Methods**:
  - Defeating monsters
  - Looting chests
  - Discovering symbols/waypoints
- **Functionality**:
  - Exclusive to maze use
  - Allows path marking to prevent player disorientation

### Progress Feedback System
- **Real-time Notifications**:
  - Monster kill counters
  - Chest opening progress
- **Reward Buffs** (on kill/open):
  - Short-duration speed boost
  - Minor HP restoration
  - Slight hunger recovery


ATT2CHTranslationTeam:
I've noticed that mobs in the maze are all spawned at the beginning... This causes explosive issues with the new health bar system - entities instantly multiply, especially in Maze-7 where the count suddenly exceeds 900. Perhaps we should modify mob spawning to occur periodically near players instead. This would both prevent mass simultaneous entities and increase encounter frequency.

Piccomaster:
Other than creating a cinematic with a timer to make the mobs appear one by one, I don't see any other solutions.
Or temporarily disabling the health bar system in this maze...

But I have a question: haven't you implemented a security feature that would limit the radius distance at which the system considers entities?
Because that could be useful in precisely this case.

Piccomaster:
Regarding the changes made to the maze.
Are you sure the final scoring system that counts points at the end of the maze hasn't been broken? Because I saw that you commented on quite a few functions that were used.

You should also remember to note the changes in the changelog file, even though I think that after testing, some things may still change.

I think the path marking is a good idea to prevent player disorientation, but it would be good to disable it if the player chooses the hardest difficulty, or make it less obvious.

ATT2CHTranslationTeam:
These newly added detection features were implemented while I played through the maze, and they've already been tested - everything works fine

As for the relationship between mazetorch and Master difficulty... essentially it's about placing torches to mark paths... It's quite difficult to implement a 'disable' function for this, since players could pre-place torches as markers during lower difficulty settings

PR #274, opened 2025-08-17
fix minor error| add rune_bundle| add spell kill XP

ATT2CHTranslationTeam:


Piccomaster:
The ability for the player to automatically store runes in a bundle is interesting.
I added a free bundle at the beginning of the adventure for this purpose in the last release. I suppose it'll have to be removed in this case?
Otherwise, what's the point of the chair?

ATT2CHTranslationTeam:
The Rune Pouch is a standalone system... I've incorporated it into crafting recipes, requiring four runes (Ra, Ust, Nym, Ehl->rune) and a Small Pouch to synthesize. If adjustments are needed, we could also provide players with one at the start of the game. As for the chair's functionality, it's purely for sitting... and prevents monster spawns nearby when occupied. The original intent was to let players fish while seated, ensuring they won't be disturbed by monsters. Beyond that, it has little practical use - but as a 'little novelty item,' I think it's quite charming

Piccomaster:
Indeed, the chair idea is also interesting and even perhaps practical.
So perhaps the changelog should just include more details on the usefulness of certain additions like this one rather than just telling how to use them, because it sometimes lacks the precision that would be more useful in making people want to play this new version of the map.
But in any case, I'll link the entire change log once this version is finished and tested to see if there are any additions or fixes to be made.

PR #275, opened 2025-08-19
fix minor error | add special elite | add qucik_recipe system

ATT2CHTranslationTeam:


Piccomaster:
I think you made some mistakes.

You completely removed the size attributes of Super and Mega Elites, and you also removed their power (strength and resistance) for Super Elites. And you even removed their power based on the game's difficulty.

It's good to add exceptions with mutations, but you shouldn't remove these added power differences between levels, especially the size attribute.

ATT2CHTranslationTeam:
Regarding the scale attribute modifications for elite mobs - this is primarily because I want to make it a distinctive feature of 'max health Mutants'. Elite mobs with the 'max health ' ability will have their size increased to 1.5, which better matches their characteristics. Conversely, 'Speed Mutant' elites will have smaller sizes. To make their traits more visually distinctive, I've removed the scale attributes from 'Super' and 'Mega' elite mobs.

ATT2CHTranslationTeam:
Regarding the adjustments to speed and strength for Super and Mega elite mobs - since there are now more types of elite mobs, the game difficulty might become too high if we don't slightly reduce the base stats of Super and Mega elites... So I've made some new changes: 'Mutant Elites' will not overlap with Super or Mega effects. That is, if an elite mob already has Super or Mega attributes, it won't gain Mutant abilities

PR #281, opened 2025-08-30
Removed some useless double quotes("") 

ATT2CHTranslationTeam:


Piccomaster:
What ?? The quotes are no longer useful in commands ? It must have taken you a lot of time to clean all that up.

orian34:
> What ?? The quotes are no longer useful in commands ? It must have taken you a lot of time to clean all that up.

Yeah, it's not always needed for keys when it's not strings.

ATT2CHTranslationTeam:
Uh.... Actually, I used scripts to replace these structures, so it didn't take much time..

PR #286, opened 2025-09-04
update spell20 system (Many small features have been added )

ATT2CHTranslationTeam:


ATT2CHTranslationTeam:
emm...,
 new storage system features have added a lot of small features that will make it easier to use, and these are some of its new features
 - ### **Stock System Update**
    - **Opening Method**: Hold **conscience book** + `Shift` + `Right-click`
 
  - ####  **Page Functions**: Previous Page Icon
    - **Click**: Return to previous page  
    - **Shift + Click**: Quick-deposit inventory items into Stock

  - ####  **Page Functions**: Next Page Icon  
    - **Click**: Go to next page
    - **Shift + Click**: Quick-move Stock items to inventory

  - ####  **Page Functions**: Discard Icon
    - **Double-click**: Pick up nearby ground items into Stock
    - **Drop**: Throw out all items in your Stock

  - ####  **Page Functions**: Quick Loot Feature
    - Hold **conscience** + Hold **conscience book** + Press `Discard key`:
    - **First press**: Preview contents of targeted chest
    - **Second press**: Transfer all items to player (only works on opened chests)
    - 
This is a preview of the chest's display:
image
it can only preview chests that the player has opened (chests that do not generate loot cannot be previewed or further use the Quick Pickup function)

Piccomaster:
Wow, this new feature is incredibly useful ^^.
I'm so lost now seeing what we can do in the functions; it's crazy what we can do now with datapacks.
But also, you have to be a little crazy too to create this kind of content, hahaha.
It's a very good idea. But I don't know if players will understand how to use all these new functions, because most of them don't read change log files.

PR #288, opened 2025-09-13
add sidebar system/more small balance

ATT2CNTranslationTeam:
The sidebar will now display the global information of the map.
sidebar


Piccomaster:
We agree that this is an option only in consciousness book to activate and not mandatory? Because now the Dahäl no longer appears at all.
It's a good idea otherwise, but showing the appearance time of the Lost Souls and Mob Invasion is not something I really like and it greatly reduces immersion. I don't want this kind of mechanism to be known; some things have to remain subtle and organic.

However, for the rest, it adds a little more comfort to the gameplay, and that's a good addition.

ATT2CNTranslationTeam:
The information about dahal will be displayed at the top of the screen, so the score bar on the right can now show other content instead of just dahal.
As for what to display... I just created a few simple contents, and we can continue to adjust it later.

PR #289, opened 2025-09-20
fix fire ball system | more minor balance |add player protect time

ATT2CNTranslationTeam:


Piccomaster:
I see you deleted the old command summon of all wither with each class that I added recently and you replace by just one file like this for exemple : ` $execute if score level DIFFICULTY matches -1 run summon minecraft:wither_skeleton ~ ~ ~ {PersistenceRequired:1b,NoAI:1b,Tags:["LVL0","$(class)","Undead","NewInvo","Reg1","TrialMinion"],attributes:[{id:movement_speed,base:0.20}],drop_chances:{mainhand:0,offhand:0,feet:0,legs:0,chest:0,head:0,body:0,saddle:0}}`
And change the summon command by this : `execute positioned -4778 54 -5374 run function att2:summon/reg_1/withertrialminion_classx {class:CLASS4}`

I guess if you did it, it's because it works like that now. But I wonder if we couldn't optimize all existing summons like that? Because there are many summon files that would be useless if we changed it like you did. What do you think?

ATT2CNTranslationTeam:
Yes, I have such plans, though it might require using some scripts to perform the replacements, which will take some time... This optimization is just an example. When we add new content in the future, we can use this format to avoid a large amount of repetitive work.

PR #290, opened 2025-09-20
add an any function system

ATT2CNTranslationTeam:
It can execute other functions within the same function file without requiring the creation of new files. I believe this would be extremely useful, as it can to some extent implement if-else logic similar to other compilers within data packs. Most importantly, these functions execute instantaneously, making them functionally equivalent to normal mcfunction file calls. The key advantage is that it eliminates the need to create additional mcfunction files. However, it's best suited for one-time trigger function commands rather than persistent loop functions, as macro functions do incur some little performance overhead.

#example
function att2:any_function/execute {function:['give @a apple','execute as @a run say i have any apple','execute as @e[type=pig] run say i am a pig']}
image
The execution order is from left to right, and I think this feature will serve a very convenient purpose.


Piccomaster:
An excellent tip for limiting the number of functions, but indeed, if it is a bit resource-intensive, it should be used sparingly.

Piccomaster:
Oh, you were quick to optimize the summon files, great!
However, as is often the case when you do massive changement, in your last forced commit, you deleted all the additions and fixes I've made over the past two weeks.
As already mentioned, when you make changes to the code in your local folder, never forget to do a "git pull" to get the code of other team members who work on the project and only then make your commits.

ElGrenier:
sooo, with your mass edit in the spawner edit you did some little overhead, like : 

```
golem0_classx {class:CLASS9}_tower_se_mech8
```

You should fix this and other thing like that before merging this PR

Piccomaster:
There is also my modification of the `gameplay/dahal/action/spell9/obtain.mcfunction`‎ file was removed again in the last commit.

PR #293, opened 2025-09-24
Replace the click text with the trigger command. |fix minor error

ATT2CNTranslationTeam:
Ok, I used some scripts to replace all click commands with trigger instructions to set scores and then execute the corresponding commands. Now we can continue to upgrade the version.^^


Piccomaster:
Wow, you did it? You're crazy, hahaha. Do you think it'll be okay in terms of resources and performance?

Be careful, though: some commands I recently added have been removed again...

ATT2CNTranslationTeam:
Although it does have some impact on the creative process, as it requires digitizing commands for execution, which greatly reduces readability, it will indeed be much better for certain servers. Now, content like books and dialogues will no longer force players to have OP permissions.

Piccomaster:
Indeed, the security is bypassed now, so it has that advantage, but I dared to hope that when version 22 was released, Mojang would add a simple gamerule so we wouldn't have to do this.

In any case, it's a great job.

ATT2CNTranslationTeam:
> Wow, you did it? You're crazy, hahaha. Do you think it'll be okay in terms of resources and performance?


It won't add too much performance overhead, as it mainly relies on the score. ...

Piccomaster:
However, we'll need to carefully navigate the code between which score ID returns which command ^^'.

I also noticed that I'll have to go back to certain locations on the map to update certain command blocks.

ATT2CNTranslationTeam:
> I also noticed that I'll have to go back to certain locations on the map to update certain command blocks.


Only the signboard and the enchanting table need to be updated, the content is not too much.

PR #300, opened 2025-10-04
update all menu system

ATT2CNTranslationTeam:
Almost all the content from the book 'Consciousness' has been fully integrated into the new menu system, while some additional guidance has also been added.
image
image
image
image
image
image
image

But there are still many details that need to be optimized... For example, the content of this image, some things need to be updated.
image


Piccomaster:
Wow I don't even know what to say, the changes are so impressive :open_mouth:. It makes it so much better and clearer to understand.

Piccomaster:
> But there are still many details that need to be optimized... For example, the content of this image, some things need to be updated.

Yes, the quickest way is to simply remove the image from the book page. It's not much use anymore if the options are now so easy to understand.

orian34:
Amazing work yeah!  I'm glad I suggested that haha, this will make the book so much more readable.

PR #301, opened 2025-10-11
update some dialog menu | add book view system

ATT2CNTranslationTeam:
I've organized all the books. Now players can use the mouse wheel to view the entire content of a book, and I've added a menu to review unlocked books.

c6caea63-6ecb-4612-8466-96abcc66e39a
19f56803-10ff-4360-ba4e-aa775f744955


Piccomaster:
But, it's just awesome!!
Great addition and very practical.

ATT2CNTranslationTeam:
> But, it's just awesome!! Great addition and very practical.

Since all the text in this dialog is "centered", I had to use some formatting scripts to add a lot of spaces and line breaks to make it appear "left-aligned". However, even so, some of the text is still a bit messy...

Then, due to the item updates for the books, you need to enter "function att2:update/all_book" in the map, and after waiting a short while, all the books will be updated automatically.

Piccomaster:
> Then, due to the item updates for the books, you need to enter "function att2:update/all_book" in the map, and after waiting a short while, all the books will be updated automatically.

Ok, I shouldn't forget, but I think I'll go over everything on the map anyway before the final update. And I'm adding new little areas all over the map so I have to update the chunks anyway.

PR #305, opened 2025-10-21
Optimize the camera movement for all animation scenes.

ATT2CNTranslationTeam:


Piccomaster:
Changed all the cutscenes, these are potentially major bugs to fix. We'll have to test everything, but I'm eager to see the results.
For the "update/all" file change, I'll try to remember the exact locations, because going through the entire map again might take a long time.
Don't forget to update the "Changelog" file with all the changes, because there are a lot of them in the latest pull requests, and I see that you've even made it possible to skip the dialogues of other cutscenes.

PR #310, opened 2025-10-29
Update/Optimize chest particle system

ATT2CNTranslationTeam:
## Update Instructions

Based on the chest data spreadsheet you provided, 
I have made modifications and configured a Python script within it. 
Now, it only requires importing that chest data spreadsheet to automatically generate the corresponding update files, which facilitates subsequent updates.

## Chest Marker Entity Update Instructions

Enter "function att2:update/chest/summon" to summon marker entities.

(Note: During updates, you can set the chunk rendering distance to 2. This prevents slow loading speeds from causing some entities to fail to load in time.)

## Script Usage

After starting the script in VSCode, input the file path to the Excel spreadsheet containing the chest coordinates.

The script will then automatically generate all the summon commands file.


Piccomaster:
Okay, great, that way everything will be updated directly on the map.
But I have a few questions.

I looked at the code details, and I don't understand why the "start" chest update file only has about 2000 chest coordinates to teleport to their locations? Whereas in the Google Drive file, we listed nearly 4500 chests.
The marker summons in the "summon" file seem to be counted correctly.

As for the Python script, we agree it's just for saving, I don't need to run it, right?

Because that brings me to my last question: I plan to add more chests to the map, and now with all the new features, I'd like to know where I should enter their locations ? Just to be sure I don't miss anything.

ATT2CNTranslationTeam:
> I looked at the code details, and I don't understand why the "start" chest update file only has about 2000 chest coordinates to teleport to their locations? Whereas in the Google Drive file, we listed nearly 4500 chests. The marker summons in the "summon" file seem to be counted correctly.

In the command generation script, I've added a distance check between coordinates. If the distance between two coordinates is less than 20 blocks, the script will automatically skip the coordinate being compared. This helps eliminate a significant amount of redundant work.

This is primarily because some chests are very densely clustered, such as in post-boss fight reward rooms or exploration warehouses like Ryliath's.

And this TP cycle is only to ensure that all markers are assigned their corresponding color scores, so there's no need to teleport to every single chest location.


> As for the Python script, we agree it's just for saving, I don't need to run it, right?

It definitely needs to be executed, but only after all the chests have been updated.


> Because that brings me to my last question: I plan to add more chests to the map, and now with all the new features, I'd like to know where I should enter their locations ? Just to be sure I don't miss anything.

You can directly input the summon command for the marker in the summon file, 
or you can continue adding chests in that Google Sheet following the same format. 
This script detects entries like "execute in minecraft:overworld run data merge block -4984 81 -5051 {CustomName:"\"Rangement\"",LootTable:"att2:chest/reg1/c1t1"}" in the 16th column of the sheet, 
extracts the dimension and chest coordinates, and generates them accordingly. 
Just ensure the format remains consistent.

ATT2CNTranslationTeam:
OK, I've re-added some features. Score initialization for markers in range is now automatic, no longer requiring the "function att2:update/chest/start" function. Just generate the marker entities("function att2:update/chest/summon".

Piccomaster:
Ok, but... just wait, let me try to summarize the steps to make sure I’ve understood everything correctly. Because I don't want to make any mistakes and you did a great job, so I don't want to ruin it...

So, basically, you created a Python script that generates functions inside the datapack to directly update the chests and their markers on the map.
This script creates or updates the functions in the **`update/chest`** folder, using the chest database from the Excel (Sheet) file stored on Google Drive.

That way, future chest updates can be done much faster.

#### 🧭 The procedure is as follows:
1. I pull the datapack code onto my map.  
2. I open the latest version of my map in Minecraft (with only 2 chunks rendering distance), then simply launch the function **`att2:update/chest/summon`**. 
3. I'm waiting for the updates with teleportation player on the map to finish.  
4. Next, I download the chest database file from Google Drive in Excel format.  
5. I set the path to this Excel file in the Python script.  
6. Then, I run the Python script from my IDE where I develop the datapack, since the Python script will stay inside the Minecraft datapack folder.  
7. Later, if I want to add more chests, I’ll just write down their positions again in the Google Drive Sheet and repeat the procedure.


So, first question:  
- Is everything I described correct so far?

And if it is, I still have a few more questions:  
- I looked through the Python script but can’t find where exactly to define the excel file path?  
- I need to install Python on my computer (it’s been a while since I last used it... hahaha) so, which version of Python should I use — the latest one, right?

ATT2CNTranslationTeam:
> * Is everything I described correct so far?

Yes, no problem.

But, you just need to input **`function att2:update/chest/summon`** - there's no need to wait for it to finish updating (it happens instantly). That's because I've added marker score initialization in the latest section. When players approach the marker entity, their scores will update automatically, so no lengthy waiting is required anymore.

> * I looked through the Python script but can’t find where exactly to define the excel file path?

Indeed, there are some minor issues with the script. It will generate the **`summon.mcfunction`** file in the same directory where the Excel spreadsheet is located. Hmm, it shouldn't be a serious problem though. You can directly place the spreadsheet into the datapack...

> * Which version of Python should I use — the latest one, right?

I'm using Python 3.13 version.

ATT2CNTranslationTeam:
Then this should be the correct operational interface. It runs in the terminal of VSCode/IDEA - at least for me, this approach is more convenient...
image

Piccomaster:
Okay, great, I'll test that then.

Piccomaster:
So I installed everything and tested the script, and it works perfectly in my environment. Everything generates correctly; I just hope we didn't forget any chests in the database ^^.

However, I have one last question: I'm trying to understand the purpose of running the command "function att2:update/chest/summon" only once. There are a lot of different locations separated by many chunks in this function, so running the command once when I'm in the center of the map won't create the entities in the chunks far from my position...

ATT2CNTranslationTeam:
> However, I have one last question: I'm trying to understand the purpose of running the command "function att2:update/chest/summon" only once. There are a lot of different locations separated by many chunks in this function, so running the command once when I'm in the center of the map won't create the entities in the chunks far from my position...

Oh, don't worry about that. The summon command will still spawn entities even in unloaded chunks. And since every entity has a unique UUID, we don't have to worry about duplicates.

PR #313, opened 2025-10-31
Fixed some issues and made various improvements|Convert the 'run_command' from the skipped dialogue to trigger

ATT2CNTranslationTeam:


ATT2CNTranslationTeam:
Added some sprite images to the product list, making it look clearer.
image

Piccomaster:
The icons in the shops are a great idea!

However, are you sure about the calculation of total exploits point?
Because according to my database, it seemed to be up to date, unless you removed some progress or decreased some points.

I think I'll share the achievements database with you too, since I lost track of it to make sure it's up to date.

Edit : Ok nevermind, I managed to update the database.

ATT2CNTranslationTeam:
The PR I submitted adds 10 achievements for the matching_game section, which should be correct. As for the overall exploration points, I'm not entirely sure, as I haven't conducted a full achievement test yet.

PR #315, opened 2025-11-03
update spawner system /add Toolbar System

ATT2CNTranslationTeam:


ATT2CNTranslationTeam:
A short video demonstrating the functionality of the toolbar:
The video quality is somewhat poor because I didn't want to make it difficult to load... so I used a small window for the demonstration.
https://github.com/user-attachments/assets/91639edf-8a8c-4989-8c0e-3cc497fc5fb1



Updates in this PR:
## Optimized Monster Spawning System
- The **natural spawning** of monsters is now accompanied by **particle effects** and **sound effects**. This provides a clear visual and audio cue to alert nearby players of their appearance.

## New Toolbar
- A new **toolbar** with four dedicated slots has been added. Players can assign tools like "Torches" and "Rune Pouches" to these slots, allowing them to quickly trigger the tool's **specific effect**.

- **Spell Pouch**: Can hold three casting implements. When placed in the toolbar, it automatically casts the corresponding spells.

Piccomaster:
That’s an excellent system! A brilliant idea. However, this kind of tool shouldn’t be given out for free — especially not at the beginning of the game, since it’s both powerful and very handy.
I’ll have to think about that, unless you already have an idea on how it could be obtained?
Selling it in a shop for ESC could be an option too, but that would mean players could have several Spell Pouches in their toolbar, and then they’d be able to cast almost all the spells nonstop — well, assuming they have enough Dahäl, which is far from guaranteed. ^^

ATT2CNTranslationTeam:
Indeed, I haven't yet decided how players should obtain this spell pouch. 
My current idea is to unlock this item through a spell-related "daily quest."
However, players should only be able to possess one at a time... because I've only implemented the functionality for a single spell pouch and it doesn't support multiple ones. 
(though it's possible to consider increasing the capacity within the pouch... this is controlled by the maximum stack size of the spellcasting item, which can be modified very easily)
Moreover, if players were to carry multiple spell pouches, it might become overly powerful .

Piccomaster:
Yes, it shouldn't be too powerful, so one per player is perfect. After that, why not make it upgradeable? For example:
- 2 daily quests completed: unlocks the pouch with 1 slot.
- 4 daily quests completed: unlocks 2 slots.
- 6 daily quests completed: unlocks 3 slots.
- 8 daily quests completed: unlocks all 4 slots.

Or indeed controlled by the maximum stack size of the spellcasting item.

PR #317, opened 2025-11-09
fix matchinggame adv | optimize pnj talking

ATT2CNTranslationTeam:
Now, after players interact with NPCs, the NPCs will continue to look at the player for a certain period of time.


Piccomaster:
I was wondering about the chests you added in the "summon" and "miss_chest" functions. Won't there be a problem since their positions aren't saved in the Google Drive file's database? Because if I add chests to the map, I'll note them in the Google Drive file, and then generate the summons with your script. Won't that take the new positions into account?
Or, I should give you access to the file so you can add the positions yourself ?

ATT2CNTranslationTeam:
No need to worry for now, this can be handled last.

orian34:
It's probably already planned, but make sure to add all new chests/whatever secret/new spots in the secret seeker list too.

ATT2CNTranslationTeam:
> It's probably already planned, but make sure to add all new chests/whatever secret/new spots in the secret seeker list too.

Yes, marker entities are now used to record the locations of various treasure chests, ores, and symbols. In Secret Seeker, simply selecting by range will display the corresponding positions, eliminating the need for exhaustive enumeration.

orian34:
> > It's probably already planned, but make sure to add all new chests/whatever secret/new spots in the secret seeker list too.
> 
> Yes, marker entities are now used to record the locations of various treasure chests, ores, and symbols. In Secret Seeker, simply selecting by range will display the corresponding positions, eliminating the need for exhaustive enumeration.

Ooh, that's great indeed!

PR #318, opened 2025-11-11
Add monster invasion events in other dimensions.

ATT2CNTranslationTeam:
And some new mechanics have been added: 
 (mainly because I felt too many monsters would increase the game's difficulty, so I added this mechanic)
1. When players open rune chests, it triggers a series of explosions that deal damage to nearby invading monsters
2. Particle effects have been added to make them look cooler.

Here's a simple video to demonstrate the specific effects:

https://github.com/user-attachments/assets/6a75eb6a-2ca4-43ba-bc8d-89787db9aeab


Piccomaster:
Indeed, the new particle effects look really good.
But it shouldn't be too powerful, because the basic goal of invasion is to kill the mobs, not to run and rush towards the chest ^^.

ATT2CNTranslationTeam:
Yes, this actually just adds a mechanism to reduce the combat power of monster hordes... The purpose is to give players more room to maneuver when they encounter monster hordes with insufficient combat strength, rather than being completely overwhelmed by them...

PR #319, opened 2025-11-16
add some dailyquests

ATT2CNTranslationTeam:


Piccomaster:
Great job, I can't wait to fully test the map with all the new features.

ATT2CNTranslationTeam:
Actually, only a small part has been completed—so far, only ten quests have been made (with over 60 planned). Most of them are tasks designed around underutilized game elements, and each city will also feature a relatively unique new quest...
 To be honest, the scale of this has somewhat exceeded my expectations... But I'll do my best to refine the system.

Piccomaster:
Don't worry, the Runic Trials system has also exceeded my expectations in terms of the work it requires, haha... I think once we've finished that, we'll need to test, fix, and polish the code before the big update.

Actually, I'm going to create another issue to discuss this because I'll need some materials. Originally, I was planning to release this huge update at the end of 2025 to celebrate the 5th anniversary of the map's official release, and since it's taking longer than expected, I'd like to write an article about the new features, without revealing too much of course, but just to tease and keep people excited about what's coming next year and show them that the map continues to evolve.

PR #320, opened 2025-11-19
Minor Optimizations to the Loot Table

ATT2CNTranslationTeam:


ATT2CNTranslationTeam:
After a long period, I finally figured out why certain long side quests (emerald) NPC transfers fail. It's because when entities with UUIDs are killed, the entity remains for 0.5 seconds (death animation). During this time, spawning an entity with the same UUID will fail, leading to a series of issues...

For example, in multiplayer scenarios, if players are simultaneously present at both the pre-transfer and post-transfer locations of an NPC, this issue will be triggered...

Piccomaster:
Oh really?? Ah, of course, why didn't I think of that sooner haha...
Don't forget to mention it in the changelog; a bug this old finally fixed is something to celebrate ^^.

PR #322, opened 2025-11-19
update dailyquest board

ATT2CNTranslationTeam:


ATT2CNTranslationTeam:
Just enter "function att2:update/dailyquest_board" and wait a moment to update the dailyquest boards for the six cities.
 To be honest, I'm not very skilled in construction, so I made a version for you to check out first. If you think something is not appropriate, you can directly modify it; 
I probably won't make any more changes to this structure.

ATT2CNTranslationTeam:
The item-to-experience conversion system has been rebuilt, using calculation formulas to determine conversion multipliers for easier future balancing, and it now supports converting any number of items of the same quality.

PR #329, opened 2025-11-25
update horse_racing / Optimize some code

ATT2CNTranslationTeam:
Oh my, so many elements are being applied to this horse racing system... In multiplayer mode, different players see different particle effects, there are random events and predefined race routes, along with AI competitors of varying difficulty levels... It feels like my head is about to split open.
I'm using display entities to show circular areas to outline some of the race track routes... This looks absolutely spectacular, haha
image


ATT2CNTranslationTeam:
Now all NPC dialogues have an avatar icon in front of the conversation
image

Piccomaster:
> Oh my, so many elements are being applied to this horse racing system... In multiplayer mode, different players see different particle effects, there are random events and predefined race routes, along with AI competitors of varying difficulty levels... It feels like my head is about to split open. I'm using display entities to show circular areas to outline some of the race track routes... This looks absolutely spectacular, haha 

Wow, that's impressive, but are you sure it won't cause too much lag?

> Now all NPC dialogues have an avatar icon in front of the conversation

It's a great idea to have done that. It's not much, but it adds to the immersion.

ATT2CNTranslationTeam:
It should be fine. At least I haven't had any pretty bad lag when playing by myself.

ATT2CNTranslationTeam:
While using regular expressions for replacement, some formats weren't fully matched, causing errors in a few dialogues. I'll fix them quickly

PR #333, opened 2025-11-27
Update the calculation method for Dahal's recovery.

ATT2CNTranslationTeam:
Dahal's recovery now occurs every tick instead of once per second.


ATT2CNTranslationTeam:
Added the final reward for the horse racing competition "Jappeloup". When players ride this horse and press the space bar to jump, the jump height will be increased and they will gain a brief speed boost.

Actually, it's about equipping the special abilities of the horses from the horse racing competition onto this specific horse. 
Since players already obtain the elytra by the time they acquire this horse, height restrictions no longer pose much of a constraint, so there's no issue with making the horse jump higher.

https://github.com/user-attachments/assets/c7cfb8d8-f5cd-42fa-bc8b-0c75886b49c7

Piccomaster:
I like the idea of ​​the horse reward with the jumping ability, which the player will have the elytra anyway. But be careful, because the elytra doesn't allow you to jump several blocks high in enclosed spaces where the player can't get a running start or gain height. But I think it'll be fine.

ATT2CNTranslationTeam:
Indeed, perhaps I should change the prerequisite for this competition to only refresh after completing all main quests. This way, at least there's no need to worry about players skipping sections in the Nojelanth part or having to nerf the horse's jumping ability in the Nojelanth segment.

Piccomaster:
Originally, players couldn't bring horses to Nojelanthe because of the time portals. But now that they can, yes, it's going to cause problems for that dungeon. Therefore, players should only be able to unlock JappeLoup after completing Nojelanthe.

PR #336, opened 2025-11-30
add ender_chest system

ATT2CNTranslationTeam:
#### Ender Chest Functionality Update
- **Absorption State**: Right-click to open the Ender Chest and activate its **Item Absorption State**, causing it to continuously absorb nearby dropped items.
- **Release State**: Left-click to activate the **Item Release State**, causing the Ender Chest to eject stored items. When near a Recycler table, items are automatically thrown onto the Recycler table.
- **Cooldown Mechanic**: Using the Ender Chest triggers a cooldown period, with duration based on the number of items absorbed/released in that operation.
- **Expedited Cooldown**: During cooldown, players can spend Chronotons to instantly complete it. Cost is proportional to remaining cooldown time.
- The storage limit is 777 items, but players cannot view the contents of the Ender Chest in real time.

- Simply enter `function att2:update/ender_chest/start` to update all Ender Chests.


Piccomaster:
What is the purpose of the Python script?

ATT2CNTranslationTeam:
> What is the purpose of the Python script?

It's mainly about sharing some practical scripts I usually use. For example, this current script can extract the coordinates of block entities, making it easier to manage chests, command blocks, etc. It allows for quick querying of corresponding locations and some data.
Later, you can use some other scripts to process this data.

image

The main reason is that I’ve noticed some current NBT editors don’t seem to have the functionality to extract all chunk data, which isn’t very convenient when batch processing data within chunks. So, I had AI build a separate script to handle this content ^^.

By processing all chunks with this script and obtaining the coordinate data of all chests, i can generate corresponding marker creation commands, which will save a lot of time.

Piccomaster:
Oh I didn't know it was possible this way. It's very practical and super well thought out.

PR #346, opened 2025-12-08
Update the new bow and crossbow system.

ATT2CNTranslationTeam:
# Arrow System Rework / New Special Arrows
## Three new special arrows have been added (with more potentially to come in the future):

- **Explosive Arrow**: Deals area-of-effect damage around the struck enemy based on the arrow's damage.
- **Tracking Arrow**: The fired arrow automatically tracks and homes in on nearby enemies.
- **Poisoned Arrow**: Slows nearby enemies upon hit and inflicts damage over 10 seconds.

The main reason is that the current arrow system lacks engaging content, and the original potion arrows are almost useless. We can add these arrows to the shop to replace the potion arrows. Of course, these arrow effects are very powerful, so they will definitely be more expensive and can be purchased using some ESC/Rune Powder.

https://github.com/user-attachments/assets/658c9825-fb2c-441c-bd0a-93f8b6352a39


Piccomaster:
Oh, very good idea!
It's true that the special arrows sold by the alchemist aren't attractive enough.

However, just one important point: rune powder is not meant to be a currency for trading physical goods.
Runes should not become a currency equivalent to ESCs or Chronotons. It should remain a relatively isolated economic system, not heavily reliant on physical goods.

However, if it is a system allowing the enchantment of basic arrows into special arrows, then using rune powder makes perfect sense.

ATT2CNTranslationTeam:
Refactored the monster health system; the damage calculation still uses vanilla content, but monster health can now be directly modified through scoreboard values.

Most boss health bars are only displayed for players who have entered the corresponding boss room. Additionally, when facing multiple bosses, the health bar at the top of the screen will be promptly cleared once a boss is defeated.

Piccomaster:
There are so many changes in this pull request, wow. The more I read the code, the more lost I feel, haha. There's such a difference in methodology compared to the old code that I understand less and less some of the functions that have completely changed... Like the use of the trident animation in legendary items or the use of copper ingots ^^.

By the way, I saw changes to the Update All function. Should I update the map again? And is the bug concerning the Interfazer armor stand spawning in loop in the center of the map fixed?

In any case, it's very good work.

ATT2CNTranslationTeam:
> By the way, I saw changes to the Update All function. Should I update the map again? And is the bug concerning the Interfazer armor stand spawning in loop in the center of the map fixed?

Not needed for now. I'm currently re-optimizing the Mythic equipment.
The code for generating the armor stand in Interfacer has been removed. Of course, you can also manually clear the remaining Interfacer armor stand at coordinates 777 77 777.

Both "trident" and "spear" are just hand animations triggered by right-clicking an item. After version 1.21.11, "spear" was changed to the animation for real spear, while the original "spear" animation became "trident", so the modification was made.

PR #348, opened 2025-12-13
Optimize Mythic equipment, add monster health reduction display.

ATT2CNTranslationTeam:
Different number icons will be displayed when using melee/bow and arrow/spells. 
Of course, players can turn off the display of this content in the relevant settings.

https://github.com/user-attachments/assets/b5e4f817-b484-46da-b785-74c93a4bd104


Piccomaster:
That's so awesome! That's exactly how I imagined the details displayed in combat, being a big fan of H&S and ARPGs.

So yes, indeed, I hope it's configurable and can be turned off, because I've had feedback from players who prefer immersion with less information on the screen. That's not my preference, but I understand why some players want it.

PR #354, opened 2025-12-21
Add more ways to break decorated pots

ATT2CNTranslationTeam:
Add more ways to break decorated pots, 
such as: 
- Jumping on them to crush them
- Using fireballs/earthquakes to destroy them
- Simply right-clicking to break them.


Piccomaster:
Later on, I'd like to add "Exploits" to track the number of pots broken by the player. We agree that regardless of the method to break it, would it be possible to calculate this and use it for advancements?

ATT2CNTranslationTeam:
> Later on, I'd like to add "Exploits" to track the number of pots broken by the player. We agree that regardless of the method to break it, would it be possible to calculate this and use it for advancements?

Of course, this structure can be used. I’ve already prepared positions for accumulating exploits inside, so it can be implemented with just a few additions.

ATT2CNTranslationTeam:
Created a new mini-game where players can engage in a fireball duel with BobL.

Competition Rules:
1. BobL will periodically launch small, medium, and large fireballs.
2. If these fireballs reach your vicinity, they will reduce your health.
3. You can launch small, medium, and large fireballs to block the fireballs BobL launches.
4. If two fireballs collide and your fireball's level is higher than BobL's, BobL's health will decrease.
5. If two fireballs collide and your fireball's level is lower than BobL's, BobL's fireball's level will decrease.
6. The match ends when either side's health reaches zero.

https://github.com/user-attachments/assets/faefaa94-645d-47d5-a40a-a51cc95374e1

Piccomaster:
Hahaha, that's a really funny mini-game. I like this little idea, why not indeed.

PR #355, opened 2025-12-22
add seaside mobs_invasion | Add a new hostile creature, Zombie Nautilus.

ATT2CNTranslationTeam:
Add a new hostile creature, Zombie Nautilus.

Primarily distributed in sea areas.


Piccomaster:
Oh, that's a very good idea; it's true that in those areas there was a lack of possible interaction.

PR #356, opened 2025-12-23
Optimize some miscellaneous items.

ATT2CNTranslationTeam:
Now, item drops can display their names and come with a glowing outline.

image

I've made a lot of detail-oriented optimizations, like adding particle effects that match the quality color of chests when opening them, or more sound effects, and additional audio cues when discovering new hidden shulker boxes... There are so many of these changes that I can't even remember everything I've added.


Piccomaster:
What you've done is just awesome!!

However, it would be great if you could update the CHANGELOG file. I'd like the changes made to be detailed in this file; it's important for the history but also for players who are interested in the changes.

Also, regarding what I see in the picture, there's just one thing that bothers me: the soup is yellow. It's a misc item, and misc items should be white. The color yellow should be reserved exclusively for runes.

Piccomaster:
I also see that you've already started coding part of Elytra's race system.

I think I'll let you handle it entirely, as I don't think I'll have time to do everything anyway.

PR #358, opened 2025-12-26
add new spell

ATT2CNTranslationTeam:
#spell 29 Battle Companion

Summons a flying pet that fights for you
You can hold a sword-type weapon and left-click on it to replace its weapon
Different quality weapons will increase its attack damage

https://github.com/user-attachments/assets/4e7ba8ea-a05b-4134-a112-157f1b89c963

#spell33 Luminous Orb
Hmm, this is the spell effect of Luminous Orb. Based on light source function, I added some other features, such as firing lasers at nearby monsters..

https://github.com/user-attachments/assets/66c94686-e36f-4f46-ac85-d1c3c51875dc

There's another spell I call "Soul Companion" (pet type).
 It has two functions: 
One is to block a fatal hit for the player, followed by a 5-minute cooldown; 
The other is that :

Left-clicking on an Allay enters the possession state, which has a duration
You can use the "Ctrl" key to move in the direction of your line of sight. 
The "A" and "D" keys are used to adjust horizontal line of sight. 
The "W" and "S" keys are used to adjust vertical line of sight. 
The "Space" key is used for interaction. 
The "Shift" key exits the possession state. 

During this time, the player can move freely, including flying, but cannot pass through walls 
Additionally, the player can trigger exploration content by pressing the spacebar (jump key), 
such as activating Symbol points, opening chests, mining rune ores, etc. 
Loot will automatically be retrieved once the possession state ends.

https://github.com/user-attachments/assets/d3e5293f-e3ca-420b-a800-f17f0014a40f


Piccomaster:
The combat pet is an excellent idea! New spells like that will add even more variety to the Dahäl's use.

The idea of ​​the symbol becoming a kind of rotating laser turret is really well done, I love it, but its appearance is redundant with the symbols you have to find in the regions. Especially since, as you level up the spell, I suppose it can stay in place for a very long time in the last level. Wouldn't it be better to just use a different object, something more "light-like"?
I remember having some objects whose textures I created, but the idea of ​​integrating them was abandoned: look at the "netherite_ingot", for example; it has an original custom texture resembling a kind of sun that could be used for this spell.

ATT2CNTranslationTeam:
While creating the spell, the first thing that came to mind was luminescence, so I naturally leaned toward the often-encountered Symbol for inspiration.Of course, I'll make adjustments to it.

ATT2CNTranslationTeam:
OK, all three new spells have been completed. Now we should consider how to add them. 
I think Battle Companion and Glowing Orb could both be placed in Eolorion for purchase, but Soul Companion needs to be tied to main quest restrictions because its abilities are too powerful—at least requiring the completion of all main quests.

If possible, we could even design some small levels centered around this spell, which I think would be quite interesting.
It would also be great to add some explorable content centered around this feature. 
In short, I think this spell has a lot of room for expansion ^^.

Piccomaster:
Ah yes, indeed, the last spell could break a lot of systems, even those of side quests and bosses.
This spell should therefore only be obtainable if the player has completed the main quest and all side quests as well.
I think making it a reward for sale at Aramis in exchange for rune powder would be great, but only after fulfilling the necessary conditions.

Piccomaster:
But to compensate for adding two spells for sale in Eolorion, one would still need to be sold in Ryliath. Couldn't the "Luminous Orb" spell work in Ryliath? Or is it too powerful to be obtainable at the start? If so, then it should be sold for over 1000 Chrotnotons.
Otherwise, as a last resort, it could be sold for ESCs on Ryliath's black market.

ATT2CNTranslationTeam:
Of course, selling the Luminous Orb in Ryliath is also a great idea. Players can start practicing the spell early on. However, no matter what, the upgrade speed for this spell will definitely be very slow. Placing it in Ryliath allows players to get familiar with the spell sooner, which is good.

orian34:
> Of course, selling the Luminous Orb in Ryliath is also a great idea. Players can start practicing the spell early on. However, no matter what, the upgrade speed for this spell will definitely be very slow. Placing it in Ryliath allows players to get familiar with the spell sooner, which is good.

What if the upgrade for this spell is linked to the amount of symbols you found, instead of regular xp?

Piccomaster:
It's an original idea indeed, but I don't want this symbol to be too prominent in the gameplay. It's meant to be an element removed from the core of the game, as it's never actually been implemented in the lore.

ATT2CNTranslationTeam:
I have already replaced its texture, and it does look much better than the "Symbol."
image

PR #363, opened 2026-01-02
Add fireball spell firing mode switching functionality

ATT2CNTranslationTeam:
The player's title name can be displayed above their head.

image


Piccomaster:
Excellent idea once again, as long as the player has a choice it's always better.

PR #364, opened 2026-01-03
Add new monster mechanics after reaching the Master level.

ATT2CNTranslationTeam:
When players reach the Master level, all monsters will trigger a brief health protection effect when their health drops below 50%. For a certain period, they gain significant damage reduction and slowly recover a small amount of health, while their movement speed decreases slightly. Each monster can only trigger this effect once.

Of course, this effect will vary based on the player's Master level and the game's difficulty setting.

https://github.com/user-attachments/assets/85748220-22cf-44f6-a977-21860d9aaabe

Then slightly reduced the spider limit in the first phase of Umbratyanth.


Piccomaster:
Hmm, I'm not quite sure how to approach this feature. I'd have to test it. Is it at least visible to the player that entities regain health like that? Because if it's without explanation or not visually explicit, it could be frustrating for the player.

As I already told you, the MASTER level is called MASTER for a good reason: the player has mastered their power, they're more powerful, and therefore should feel it. I wouldn't want it to suddenly become incredibly difficult for no reason if they encounter even the smallest elite mob.
In short, these kinds of additions need to be implemented carefully and with balance.

ATT2CNTranslationTeam:
> Hmm, I'm not quite sure how to approach this feature. I'd have to test it. Is it at least visible to the player that entities regain health like that? Because if it's without explanation or not visually explicit, it could be frustrating for the player.

Players can see that when monsters recover health, there will be a noticeable sound cue, particle effects, and a prompt similar to the damage display—showing an icon along with a number.

Piccomaster:
Okay, but I've been thinking and I've come up with an idea to prevent it from causing too many problems if the player progresses too far in leveling before finishing the story.

So I'd like another condition for this system to trigger. In addition to reaching Master levels, this system should only work if the player has completed the main story.
Because the endgame content needs to unlock at that point, especially.

ATT2CNTranslationTeam:
A portion of the elytra race has been organized. 
If you want to test it, you need to enter `execute in minecraft:overworld run tp @s -5727.17 19.00 -4577.21 -203.94 79.79`, then `function att2:gameplay/elytra_racing/initialize`. 
After that, clear the top layer of "dirt piles," stand within the range of the "air vent," hold shift for 3 seconds, and release to automatically launch. 
The remaining rules are as I introduced in the issue over there.

Piccomaster:
Okay, I'll test Elytra's race to see how it goes so we can agree on the system.

Regarding the video of the monster regaining health, for players in Master difficulty, isn't the effect a bit overpowered? Does the monster regain 50% of its health? I hope that's not the case from Master 1 level.

ATT2CNTranslationTeam:
The amount of health restored is a fixed value that varies depending on the game difficulty.

Piccomaster:
Okay, but in that case, you'd need to detail what this new mob health regeneration system does. But perhaps you already planned to include it in the README file? In any case, I don't see where it's mentioned at the moment.
What you're showing in the video... what level of difficulty is it?

ATT2CNTranslationTeam:
The difficulty in my video is Nightmare, and the player level has reached Master level 100+

ATT2CNTranslationTeam:
Hmm, editing the route is a bit complex. 
First, in the file "adventquest_function\data\att2\function\gameplay\elytra_racing\go.mcfunction," you need to uncomment the line `#execute as @n[type=armor_stand,tag=circle,tag=showing,tag=following] at @s run function att2:gameplay/elytra_racing/edit_route/circle_show`, so it runs in a loop. 

Enter `data modify storage att2:elytra_racing route set value []` to reset the route recording data. This should be done before starting the recording.


Then, enter `function att2:gameplay/elytra_racing/edit_route/give_trigger` to get the route editor (right-click to record route data). 

The first right-click will display a ring with a radius of 5 meters. 

The second right-click will keep that ring in place (while recording its position) and generate another ring, and so on. 

Once you've recorded the route, enter `function att2:gameplay/elytra_racing/edit_route/get_data` to get an item. 

Place this item in any chest, then press F3+I while looking at the chest to copy its data. 

Next, open and run the route editing script, paste the copied data into it, and also enter a custom tag name along with a starting UUID number based on the existing marker count. 

For example, if there are already 100 elytra route markers, you'd input 101. 

After entering, the script will automatically generate a file—this file is what players will call when starting the challenge.

The file should be placed in the `adventquest_function\data\att2\function\gameplay\elytra_racing\start\` directory.

To clear the displayed route entities, you can enter `kill @e[tag=circle]`, or uncomment `#kill @e[tag=circle]` in `adventquest_function\data\att2\function\gameplay\elytra_racing\go.mcfunction`.

PR #376, opened 2026-01-21
add new loottable

ATT2CNTranslationTeam:


Piccomaster:
Indeed, the optimization is impressive, assuming it still works, of course. ^^
In any case, it will save a significant amount of RAM required to run the map, as the loot tables were very resource-intensive at startup.

Afterwards, I'd appreciate it if you could add a comment to one of the main files of this loot system (for example, in this one: "gameplay/leveling/monster/loot/setdrop.mcfunction") to explain how all this system works.
Even if there are a lot of comment lines at the beginning of the file, it's not a problem; having more context is always better for those who want to understand.
This should also be done for other fairly complex systems, like the new menu, as it would greatly help me and others to understand and correct the code if necessary.

Edit: Not to mention the changelog file which also needs updating, since there have been some changes now between recent balancing and modifications in gameplay and even optimization, it should be specified.

ATT2CNTranslationTeam:
I added some explanations. In short, this involves extracting the item drop data from the spreadsheet in advance, then using the datapack to detect c/t values and classlevel/gamelevel to assign specific probabilities. After that, the loot table directly checks the corresponding scores and draws items. Thanks to this calculation method, if you want, we could even implement that the player's luck stat increases the chance of dropping higher-quality items.

By the way, optimizing the loot tables using this method reduced the total file size by over 100 MB (uncompressed).

Piccomaster:
Great, thanks for the details.
Regarding the implementation of the LUC stat to increase the chances of obtaining better quality items, it would indeed now be possible and a good idea.

However, we'll have to be extremely careful with balancing because it will greatly affect the gameplay and difficulty of the game if the player has more or fewer powerful items. And it will make the LUC stat even more powerful, so we'll also see the evolution of the number of Skill points required from the player to unlock each LUC level...

orian34:
> Great, thanks for the details. Regarding the implementation of the LUC stat to increase the chances of obtaining better quality items, it would indeed now be possible and a good idea.
> 
> However, we'll have to be extremely careful with balancing because it will greatly affect the gameplay and difficulty of the game if the player has more or fewer powerful items. And it will make the LUC stat even more powerful, so we'll also see the evolution of the number of Skill points required from the player to unlock each LUC level...

What if LUC stat is not possible to increase through skill points? What if there's specific quests/main quest progression that reward you with an extra LUC stat? That way we have absolute control on the LUC progression, and it means you don't end up with gameplay scenarios where you kill off your combat prowess for loot rate when spending level up stats, creating unbalance.

I always disliked games that had luck as a stat, making you choose either have shit gameplay but get more stuff, or have regular stats but promises of more loot is just out of reach.

Piccomaster:
> What if LUC stat is not possible to increase through skill points? What if there's specific quests/main quest progression that reward you with an extra LUC stat? That way we have absolute control on the LUC progression, and it means you don't end up with gameplay scenarios where you kill off your combat prowess for loot rate when spending level up stats, creating unbalance.
> 
> I always disliked games that had luck as a stat, making you choose either have shit gameplay but get more stuff, or have regular stats but promises of more loot is just out of reach.

Interesting point, I've never thought about that... Personally, I've always been used to hack-and-slash games where increasing your character's item drop rate was a core gameplay element.

It's a philosophy worth discussing, and I completely understand that some people don't like it, but our core game was already designed as a hack-and-slash like Diablo or looter-focused FPS like Borderlands. Changing this philosophy, which is already implemented on the map, goes too far against what ATT2 is all about.

I accept minor changes, but this one seems too significant, especially since the idea of ​​increasing LUC based on quest progress aligns with increasing loot rarity based on monster level and therefore character level.
This would result in an almost exponential effect towards the end of the game, making it extremely difficult to balance everything.

Loot is a central aspect of the gameplay in ATT2, and increasing LUC was already designed as a very useful ability. That's why it's so expensive to level up and rare on items. If we increase LUC across all loot tables, we'll inevitably have to make it even more expensive for each player to level up.

For the moment I don't see any other options. Or do you have other ideas?

ATT2CNTranslationTeam:
I actually quite like the luck stat. I often invest a bit more in luck and movement speed at the start, then go around opening chests for resources (since I almost always play on Nightmare difficulty, I tend to avoid fighting monsters in the early game).

orian34:
> Loot is a central aspect of the gameplay in ATT2, and increasing LUC was already designed as a very useful ability. That's why it's so expensive to level up and rare on items. If we increase LUC across all loot tables, we'll inevitably have to make it even more expensive for each player to level up.
> 
> For the moment I don't see any other options. Or do you have other ideas?

Well, as you said, it was designed that way, so changing how it works might be difficult since it touches the entire map's systems.
Guess it was more of a personal preference I shared, but it doesn't seem realistic to implement, at least not that way here.

What if it used more the runeword system?(keeping it separate from main progression) Or maybe it already does, I don't remember.

But yeah, my only point was mostly that I tend to prefer having luck increases be separate from where we spend level up points.

Piccomaster:
Indeed, the most logical solution would be to transfer the LUC bonus to a new LUC runeword bonus. However, this would require far too much balancing of the player's stats, as they would then have more skill points to allocate.
Edit : Or, to avoid too drastic a rebalancing of LUC skill points, could we replace the LUC stat with another stat? But what other bonus could we use? There are so many already ^^.

Furthermore, it would necessitate creating even more runewords, with a more balanced distribution of runeword bonuses if LUC were to be included, because it would be too significant a runeword bonus and would unbalance all existing runewords.

I'm not against gameplay changes, especially in an update like this, but I'm afraid of altering something too significant and difficult to balance. In any case, for me, this option would be the most consistent if LUC were to be decoupled from the player's level.
I might consider it.

ATT2CNTranslationTeam:
But if runewords are used to enhance the player's luck attribute, then how should players obtain these runewords? If players don't have enough luck, acquiring high-quality runes could become a challenge...

orian34:
> But if runewords are used to enhance the player's luck attribute, then how should players obtain these runewords? If players don't have enough luck, acquiring high-quality runes could become a challenge...

It would be gradual runewords, with increasing tiers so you don't need extreme luck to get the first few increases.

> Edit : Or, to avoid too drastic a rebalancing of LUC skill points, could we replace the LUC stat with another stat? But what other bonus could we use? There are so many already ^^.

Haha, true, and it's hard to add even more power increases. Mobility is also a nogo.

What if we added a SCAvenger stat? It would give you 0.5 food and 1 health(per stat level) when killing an enemy (with 3-5 seconds cooldown).

ATT2CNTranslationTeam:
> What if we added a SCAvenger stat? It would give you 0.5 food and 1 health(per stat level) when killing an enemy (with 3-5 seconds cooldown).

This seems more like a passive skill or something special weapons should have

ATT2CNTranslationTeam:
However, I still hope to retain the existing luck attribute system and add other ways to enhance luck. For example, introducing an enchantment in the enchanting system that grants extra luck for the next kill after defeating three monsters (there’s already an enchantment that boosts luck for the next few chests after opening multiple chests). Alternatively, we could set up special monsters that drop more loot when defeated, or build on the existing combo kill system, where consecutive kills improve the quality of loot from subsequent monsters. Since many systems are already designed around luck—such as mimic chests, boss rewards, artifacts, and newly added enchantments—making changes would require extensive adjustments.

orian34:
> However, I still hope to retain the existing luck attribute system

Regardless of any change, Luck itself would still function as it does. The question is purely about the means of obtaining it.

Piccomaster:
There are already elite monsters which drop many more items, therefore potentially more chance of having rare items. Bosses are the same thing.
Ultimately, I think adding even more chances to drop rare items isn't necessary, as there's already a high chance of items dropping from monsters, especially at higher levels.

It's important to understand that we didn't originally plan for monsters to drop so many items. Minecraft's loot system was very limited before, which is why the LUC stat is also very limited and why there are no loot enchantments on weapons.

We always felt that from the end of the game onward, when the player was level 50 and killing class 16-20 monsters, the drops were almost doubled and therefore far too high compared to the beginning of the game. We couldn't implement a linear progression for the drops, and the algorithm we created for loot is based on the limitations of the old loot tables with "rolls," "bonus rolls," "weight," and "luck."

Therefore, we didn't overuse the luck stat, because otherwise the game would have been too easy. In fact, the tests that took us the most time were precisely those related to loot, because at least 75% of the game's difficulty relies on the power of the items. We had to restart the map many times from beginning to end to evaluate the loot and balance it in order to have a worthwhile difficulty progression. Modifying the LUC system would inevitably impact this and would require extensive testing of the entire map.
The random loot system is extremely addictive, but because it's random, balancing it requires a huge amount of testing.

ATT2CNTranslationTeam:
I have conducted some tests on the latest loot table system, and it runs without any issues. However, some of the previously used loot table systems still need to be updated because they heavily reference chest loot tables, so equivalent replacements are required.

PR #377, opened 2026-01-26
add block mimic

ATT2CNTranslationTeam:
https://github.com/user-attachments/assets/66b3cda4-f12f-461b-a35e-1a72a71f9d76

A new special creature, the Block Mimic, can disguise itself by mimicking nearby blocks and hide anywhere in the world. Players need to left-click attack them to awaken them. Once awakened, the Block Mimic will randomly either attack the player or flee. Killing them grants players additional loot.

### DIFF
--- a/adventquest_function/data/att2/function/gameplay/enveffect/mimic/block/armor_stand_go.mcfunction
+++ b/adventquest_function/data/att2/function/gameplay/enveffect/mimic/block/armor_stand_go.mcfunction
@@ -1,0 +1,11 @@
+#################################################################
+#Made by Adventquest											#
+#armorstand marker go                                           #
+#################################################################
+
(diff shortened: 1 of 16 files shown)


Piccomaster:
I like the idea, but will it involve manually defining locations to place these mimics? Because there are some rather empty areas in Telluron's dimension that could accommodate these kinds of little surprises.

ATT2CNTranslationTeam:
No need to manually place these points. They are randomly selected based on existing monster spawn points. The system periodically chooses some spawn point locations near the player at random, then fires arrows in random directions around them. The block mimic is generated only when an arrow hits a block.

Piccomaster:
It's an ingenious system ^^.

PR #378, opened 2026-01-29
Add bat chest loot table drop function.

ATT2CNTranslationTeam:
`function att2:summon/bat_chest_loot_experience_bottle {C:2,T:2,Reg:Reg1,Time:10}` can summon a bat that carries a chest-type loot table, with quality tier c2t2 and drops items from reg1, generating loot after a 10-tick delay.


Piccomaster:
Wow there are a lot of changes in this pull request ^^.

I had two questions about some of the changes: why do you put a "#" symbol on most scoreboards like this:
```
scoreboard players set temp_value_13 CAL 0
scoreboard players set #temp_value_13 CAL 0
```

And also, why are the spawners like here "spawner/reg1/adanoi/catacomb/undead0_class5_1653.mcfunction" now marked as armor stand spawns?
`summon minecraft:armor_stand -3967 79 -5555 {Marker:1,Invulnerable:1b,Invisible:1b,Tags:["SpawnerAnimation"],equipment:{head:{id:"diamond_helmet",components:{enchantments:{"att2_enchantm ent:tick/misc/spawner_animation":1},attribute_modifiers:[],unbreakable:{},item_model:"nothing",equippable:{slot:head}}}},data:{summon:"execute positioned -3967 79 -5555 run function att2:summon/reg_1/undead0_classx {class:CLASS5}"}}`

ATT2CNTranslationTeam:
`scoreboard players set #temp_value_13 CAL 0` 
Adding `#` at the front is part of the performance optimization mentioned earlier. 
Including `# `means the scoreboard won’t first check for player names, reducing a small amount of performance overhead. 
The armor stands in monster spawning are mainly due to the difference between the coordinates used for generating monsters and the detection positions of spawn points.
 If markers were directly used for monster spawning, it could lead to positioning errors. 
Therefore, I generate an armor stand at the monster spawn location with an enchantment that triggers a specific monster spawning function, which also helps reduce the performance impact of `@e`.

Piccomaster:
Ok thank you, for these explanations, I had indeed forgotten certain points.

PR #379, opened 2026-01-31
update slots loot

ATT2CNTranslationTeam:


Piccomaster:
Why did you change all dye with the same color ? 🤔
Capture d&#39;écran 2026-02-01 092958

ATT2CNTranslationTeam:
Because this system mainly uses `custom_data` for differentiation, I replaced them all with the same item.

PR #382, opened 2026-02-03
fixed various syntax bugs

ElGrenier:
Soo, my IDE gave me quite a lot of random syntax error, so i permited myself to like... fix it, please check througly my review to be sure i did not fixed it badly

+ Some change to the music battle timer so its don't stop at 3 minutes instead of 4.


ElGrenier:
Also please note that its more optimized to do the TYPE check after some tags checks etc (its more costly to do a type checks than tag check, but less than the horrendous nbt check)

PR #385, opened 2026-02-12
update elytra_racing

ATT2CNTranslationTeam:
Build route editing functionality within the game. Now we can directly edit routes in-game, allowing us to add, delete, modify, and query route checkpoints, as well as create new routes.

I can only ensure that all elytra race routes cover nearly every area of the map, but they may not be the most "perfect" routes. Moreover, we may not have sufficient time to further refine this system. Therefore, I’d like to make the elytra race route editing process available to players—for example, after completing all the basic flight routes on the map, players would obtain the editing tools for this system. This way, players can create their own favorite flight routes, allowing the map to showcase its full potential. Additionally, this feature will also make it more convenient for us to edit routes ourselves.

https://github.com/user-attachments/assets/df6a4d33-28e2-42b7-b5e6-3b59ee3290c7


Piccomaster:
Wow, it's an impressive tool and really practical, but giving it to the players I'm not sure it's a good idea. They're not supposed to have such power at their disposal, or maybe it's given as an endgame reward very late in the game, because I don't know if completing all the Elytra routes is that complex.

Anyway, I'm really looking forward to trying all of this out.

PR #392, opened 2026-02-20
update enchantment

ATT2CNTranslationTeam:
You can enter `function att2:gameplay/elytra_racing/initialize` to initialize all the elytra race data. 
After that, you just need to equip your elytra and clear the obstacles at the takeoff point to start experiencing the content.


Piccomaster:
I didn't quite understand; do I need to initialize this function in the original map I'm currently building in for the release version?

Also, this file is removed from the pull request. Does that mean spears can't be enchanted?
`adventquest_function/data/att2_enchantment/tags/item/enchantable/spear.json`

ATT2CNTranslationTeam:
> I didn't quite understand; do I need to initialize this function in the original map I'm currently building in for the release version?

This function will be triggered during game initialization. However, if you need to test it early, you can directly input this function to start testing. At the same time, if you need to restore the initial state, simply input this function again to reset its state.

> Also, this file is removed from the pull request. Does that mean spears can't be enchanted? `adventquest_function/data/att2_enchantment/tags/item/enchantable/spear.json`

 As for the deleted content, it's just some unused data that you don't need to worry about.

PR #396, opened 2026-03-03
Add special mechanics for some weapons.

ATT2CNTranslationTeam:
I created a brand new mechanic for the axe. When the player holds the axe and presses the "F" key (the key for swapping items between main hand and off-hand), the axe will be thrown. The thrown axe will bounce once, and upon hitting a block or any target, it will deal damage to nearby enemies. (The axe only deals 70% of the player's base attack damage. Additionally, each bounce or contact with a block will reduce its durability and also decrease its damage by another 70%.)

**axe**

https://github.com/user-attachments/assets/0adc1483-8cd3-4737-8340-8c295fec8ab1

**hammer**

https://github.com/user-attachments/assets/e67a85a2-75c9-4d70-9429-0366c990f820


Piccomaster:
I'm impressed by the axe throwing effect, it's really cool!

PR #399, opened 2026-03-06
Add monster shield functionality.

ATT2CNTranslationTeam:
When a player deals damage to a "shield-wielding enemy" within a 4-meter range in front of them, the melee damage received by the "shield-wielding enemy" is reduced by 70%. Using different weapons can reduce the shield durability of the "shield-wielding enemy," with axes significantly decreasing the shield durability of the "shield-wielding enemy."

### DIFF
--- a/adventquest_function/data/att2/function/gameplay/enemy_health/kill.mcfunction
+++ b/adventquest_function/data/att2/function/gameplay/enemy_health/kill.mcfunction
@@ -3,6 +3,9 @@
 #detect enemy health reduce                                     #
 #################################################################
 
+
+##add tag
+tag @s add killed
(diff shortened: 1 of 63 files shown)


Piccomaster:
Interesting gameplay addition, I like it.

However, I don't know if this has already been done, but I think we should expand the "tips" message section to include all the gameplay additions in this new version, because there are sometimes subtle additions that players might not notice on their own, and there are so many now, haha.

ATT2CNTranslationTeam:
I organized it once before, but it's clear that the previous version is already outdated... Anyway, I'll wait until all the content is updated before organizing the tooltips again.

PR #402, opened 2026-03-15
update loot table 26.3.15/add new weapon mechanics /update change log 26.3.16

ATT2CNTranslationTeam:
Run the script, enter the file path of "macro loot table (new).xlsm" in Google Sheets, and execute it. 
After that, two folders will be output: one is the "entities" folder, and the other is the "chest" folder, which contain the corresponding loot tables respectively. 
You just need to copy these two folders into the "loottable" folder and replace the original ones.


ATT2CNTranslationTeam:
##spear
- When attacking with a spear, if you are too close to the enemy, the attack damage is reduced by 50% to 70%. If the distance to the enemy is just right, the attack damage is increased by 10% to 50%.


https://github.com/user-attachments/assets/6abfcb3d-5b04-4dcb-9e46-5aaa0ed4c841



##scythe

- When holding a scythe, you can see "weak points" on enemies. Aiming the crosshair at an enemy's weak point during an attack increases melee damage by 40% to 60%. If you do not hit the weak point, damage is reduced by 20% to 30%.

https://github.com/user-attachments/assets/85bdcbaa-7bff-4092-afa7-c47969c580f5

Piccomaster:
Great new mechanics, both are really well thought out for these types of weapon.

Piccomaster:
However, you overwrote the latest changes I made regarding loot... I'll wait until you reinstate my latest commits before accepting. [gameplay/leveling/monster/loot/setdrop.mcfunction‎](https://github.com/Piccomaster/AcrossTheTime2/pull/402/changes#diff-f100fb9a49b066993dbe1bc5763b05be74f5f931ee47f78283bb8f34bbba3bcd)

ATT2CNTranslationTeam:
OK, I've fixed the data.

PR #403, opened 2026-03-19
Update the calculation method for STR/HER/RES.

ATT2CNTranslationTeam:
The Strength attribute calculation formula is now synchronized with the formula in the table.

RES: Now, RES can provide up to 80% damage reduction. After exceeding 8 points, each additional point provides 5 points of fixed damage reduction.

HER: Now, each HER point provides 0.05 HP/second health regeneration. 
And instead of using the potion effect's regeneration to provide health restoration, an accumulation method is used.
For example, if HER is 1 point, 0.05 HP is accumulated per second. 
After 20 seconds, the accumulated value reaches 1 point, and 1 HP is restored.
When the accumulated value becomes negative, 1 HP is deducted.


Piccomaster:
Okay, we shouldn't forget to mention these changes in the changelog file, explaining them without going into too much detail so it's clear for everyone, because this also touches on the core gameplay. Anyway, it's much better this way, and I also saw that you did the same thing for the HUN.

PR #404, opened 2026-03-21
Add a new attribute (CRT, Critical hit coefficient.)

ATT2CNTranslationTeam:
Different weapon types have different critical hit multipliers. 

For example, an axe has a base critical multiplier of 0.02, while a dagger has a base critical multiplier of 0.05. 

If the player's CRT attribute is 5, then when attacking with an axe, the critical hit probability is 10% (0.02 × 5), and the damage dealt on a critical hit is increased by 110% (0.02 × 5 + 1). 

When attacking with a dagger, the critical hit probability is 25% (0.05 × 5), and the damage dealt on a critical hit is increased by 125% (0.05 × 5 + 1).


Piccomaster:
That's a really interesting addition. So, if it's a completely new attribute, is it possible to apply it as a bonus or penalty to weapons and armor, and possibly to allocate points using skill points earned through leveling up?

ATT2CNTranslationTeam:
Of course. Actually, this critical hit rate is meant to compensate for weapons with low damage. However, this attribute should be quite rare, because if it becomes too common, the overall strength will increase significantly. Alternatively, the related multiplier could be lowered to maintain balance. Currently, I’m positioning this attribute similarly to the RES attribute—both are very powerful stats.

PR #411, opened 2026-04-07
update new temperature system

ATT2CNTranslationTeam:
Players now have a "Body Temperature" attribute, which is influenced by various events and actions, such as being in water, sprinting, rain, wearing different types of armor, or even casting fire spells. These factors can increase or decrease body temperature. Additionally, different regions have an "Ambient Temperature" that directly affects the player's body temperature, causing a fixed amount of temperature loss per second.

When the player's body temperature is between -100 and 100, they gain a +1 bonus to SPD and HAS attributes. 
When body temperature drops below -200, a Cold Penalty is triggered, applying negative HER and HUN effects based on the current ambient temperature for 5 seconds, along with a 3-second Poison effect. 
When body temperature exceeds 200, a Heat Penalty is triggered, applying negative SPD and HAS effects based on the current ambient temperature for 5 seconds, along with a 3-second Wither effect.

This new temperature system provides players with both positive and negative benefits, with punishment and reward cycles being more gradual rather than constant negative or positive effects. 
Additionally, players have more ways to manage their body temperature.


orian34:
Ooh, that's a cool idea! I did feel like the temperature mechanic was in need of a glow up

Piccomaster:
Yeah it is an excellent idea to rework the temperature system. It was an addition that lacked depth.

PR #412, opened 2026-04-07
Add a function for spellcasting to affect temperature, and refactor the random logic of the Assist system.

ATT2CNTranslationTeam:
Now, when randomizing, the Assist system will randomly draw from all available tips and will not repeat the same tip until all tips have been displayed.

I noticed that I seemed to have deleted the decorated_pot.json file from the loot tables... but I've since recovered it. Hopefully, that won’t cause any issues.

Also, there's a small issue: the loot logic for decorated pots is that they only drop one item—meaning the entire loot table only generates a single item, even if multiple rolls are made. I’m not sure if this was taken into account during the design, so I wanted to ask: if you want them to generate multiple items like chests do, I can make the necessary adjustments.


Piccomaster:
Indeed, I hadn't noticed that the loot table for the decorated pots had been removed. I'm actually placing the pots on the map right now.

I didn't think there would only be one item per destroyed pot that would drop from the loot table. Yes, it should behave like the chests, because I put several Rolls in the loot table, the one for runes and the one for currency.

Edit : However, if it's too complicated and requires a makeshift system with resource-intensive entities, we'll leave it as is; it won't be a big deal.

ATT2CNTranslationTeam:
There won't be much performance impact, so no need to worry about that

Also, have you recorded the positions of these decorated pots? The "Seek Secret" spell requires that location data.

Piccomaster:
No, I haven't noted any positions; there are too many. It's just bonus loot; the goal isn't to know all the locations to loot them all. Progress based on the number of pots broken will be approximate.

PR #413, opened 2026-04-08
add finaljudgment/ update pot loot detection

ATT2CNTranslationTeam:
- The decorated pots now drop their loot normally instead of dropping only a single item.

- I've upgraded the visual presentation of this flying giant scythe. Now they look even cooler.

Each time the giant scythe hits an enemy, it grows larger and deals increased damage. After being launched, the giant scythe will first move forward slowly, then drift toward nearby enemies, and finally return to the player. When the player comes into contact with the returning scythe, it will be retrieved, granting a brief period of invincibility and dealing damage to nearby enemies.
Additionally, the giant scythe will attack the enemy's "weak points" within a relatively thin planar range. Successful hits will also result in increased damage.

https://github.com/user-attachments/assets/c7818d9f-492e-4cd5-921f-7d32e9e1ff37


Piccomaster:
Great effect!! Now I'll have to find a specific skin to make for this mythical scythe.

PR #415, opened 2026-04-09
Convert all boss bar texts into translation keys.

ATT2CNTranslationTeam:
- Modify the boss's title display so that it triggers every time the player enters the boss door, rather than only triggering once at a specific phase.
-  Convert all boss bar texts into translation keys.
- Fixed a bug: when players were fighting a boss inside the boss room and died, if they re-entered the boss room within 0.5 seconds of their death, the boss would be directly marked as dead. The main reason is the previously mentioned issue—after an entity dies, there is a 0.5-second death animation. During this period, an entity with the same UUID cannot be spawned, causing the boss to be incorrectly determined as already dead.So now, whether the player fails or wins in a boss fight, the boss room will enter a 2-second cooldown, during which the related boss logic will not run, preventing any judgment errors.

### DIFF
--- a/adventquest_function/data/att2/dialog/consciousness/guide/menu.json
+++ b/adventquest_function/data/att2/dialog/consciousness/guide/menu.json
@@ -38,20 +38,6 @@
                 "type": "run_command",
                 "command": "/trigger ScoreTrigger set 2859"
             }
-        },
-        {
-            "width": 150,
-            "label": {
-                "translate": "consciousness.menu.equipment_swap",
-                "color": "#21BCFF"
-            },
-            "tooltip": {
-                "translate": "consciousness.menu.equipment_swap.tip"
-            },
-            "action": {
-                "type": "run_command",
-                "command": "/trigger ScoreTrigger set 2859"
-            }
         }
     ],
     "exit_action": {
--- a/adventquest_function/data/att2/dialog/gametip/attribute.json
+++ b/adventquest_function/data/att2/dialog/gametip/attribute.json
@@ -1,196 +1,210 @@
 {
-    "title": "",
-    "type": "minecraft:notice",
-    "pause": false,
-    "columns": 1,
-    "body": [
+  "title": "",
+  "type": "minecraft:notice",
+  "pause": false,
+  "columns": 1,
+  "body": [
+    {
+      "type": "plain_message",
+      "contents": {
+        "translate": "consciousness.gametip.attribute",
+        "color": "green"
+      }
+    },
+    {
+      "type": "plain_message",
+      "width": 400,
+      "contents": [
         {
-            "type": "plain_message",
-            "contents": {
-                "translate": "consciousness.gametip.attribute",
-                "color": "green"
-            }
+          "text": "1.",
+          "color": "#FA8C16"
         },
         {
-            "type": "plain_message",
-            "width": 400,
-            "contents": [
-                {
-                    "text": "1.",
-                    "color": "#FA8C16"
-                },
-                {
-                    "translate": "att2.assist.attribute.1",
-                    "color": "#FA8C16"
-                }
-            ]
-        },
-        {
-            "type": "plain_message",
-            "width": 400,
-            "contents": [
-                {
-                    "text": "2.",
-                    "color": "#FA8C16"
-                },
-                {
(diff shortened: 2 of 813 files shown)


Piccomaster:
Great, but why did you delete all the json files corresponding to the mending in the folder `adventquest_item_modifier\data\att2\item_modifier` ?

ATT2CNTranslationTeam:
Since the durability repair is now handled through more convenient function-based calculations rather than exhaustive enumeration, these files are no longer needed.

Also, regarding translation keys: by default, if the selected language does not have a corresponding translation key, it will fall back to using the en_us translation key. Therefore, you don't need to add too many duplicate English JSON files.

Piccomaster:
> Since the durability repair is now handled through more convenient function-based calculations rather than exhaustive enumeration, these files are no longer needed.
> 
> Also, regarding translation keys: by default, if the selected language does not have a corresponding translation key, it will fall back to using the en_us translation key. Therefore, you don't need to add too many duplicate English JSON files.

Oh, ok great ! So I will do some cleaning then in the resource pack.

ATT2CNTranslationTeam:
Regarding a small new exploration feature, I think we could add some weapon rewards at specific locations. 
image

For example, at this location with the iron sword, we could create a simple `interaction entity` + `display entity` structure and place it at a specific spot on the map. 
When players pick it up, they would need to meet certain attribute requirements—such as having at least 5 Strength points to be able to lift the sword. 
Upon pickup, the weapon would be randomly generated within a specified rarity range, such as among the UNC, RAR, and EPI rarities. 
I think this kind of random content would be more interesting than adding more chests. 

Plus, I feel there are already too many chests in the game, which can easily lead to fatigue.

By the way, I plan to organize all entity positions on the map (especially NPCs) and replace all armor stands with display entities. Display entities allow setting a specified rendering range, while armor stands do not, and armor stands put more pressure on rendering than display entities. This will help reduce rendering stress. Additionally, I will organize the positions of all entities/NPCs to prevent irreversible losses caused by players using the kill @e command.

Piccomaster:
Why not, indeed that would be a good idea. However, I completely disagree with the idea that there would be "too many" chests. In my opinion, there can never be too many chests in an RPG. Especially in the open areas of Sylberländ, which are very, very lacking in interest in some places. I simply plan to add chests and areas with decorated pots in some of the places I consider empty. That's also why we shouldn't add too many rewards on top of the chests; it's true that since the introduction of mimics, there are too many rewards, which is why we should reduce their spawn rate for low-level chests.

Regarding the ground equipment that you pick up and that gives you a random weapon, I'm not really in favor of that option; it's strange and would complicate things considerably. However, placing Chronotons, runes, and ESCs in specific locations on the ground that the player could pick up would be pretty cool.

ATT2CNTranslationTeam:
Of course, other items can be placed as well. This feature is actually very simple to implement.

As for chests: my idea is to place some chests in more hidden locations rather than out in the open, because putting chests in wide-open spaces feels a bit strange. I think using these pickable items to fill areas that lack elements would be better than filling them with chests.

orian34:
Reminds me of the datapack (made by someone else) I was using while making ctm maps

image

Piccomaster:
First of all, I don't fill empty spaces with chests randomly; I carefully place them in specific locations, hiding them or slightly modifying the terrain. I don't just throw thousands of them around for no reason either.
During map development, we never did that; there are very few chests lying around, so I don't understand the concern.
Especially since many excellent RPGs, and even recent ones like Crimson Desert, do it, and I don't see the problem.

Furthermore, it won't prevent you from placing items on the ground or in specific contexts as you wish. I even think it would add more interest to exploration. But if you want to work on that, you'll have to wait until I finish my additions first, and only then will I give you the map, because otherwise there might be a conflict in the positions of the areas of interest.

ATT2CNTranslationTeam:
However, I won’t just leave these items out in the open on the map. Instead, I’ll bury them like this.

https://github.com/user-attachments/assets/72d4d2f1-a163-4e63-9fd6-6fdfd4be1b9f

Piccomaster:
Excellent, I really like the concept.

ATT2CNTranslationTeam:
image

I created 7 overlay textures corresponding to some simple structural locations, and added a location data collection table and test commands in the Google Sheet.

PR #416, opened 2026-04-12
update finaljudgement

ATT2CNTranslationTeam:
Added a left-click effect for the mythical scythe, creating a simple interaction with the right-click thrown giant scythe. Left-clicking the giant scythe as it returns to the player will cause it to fly forward again slowly.

Although there were plans to add more effects, I think this mythical is already interesting enough as it is and doesn’t need any additional features.

https://github.com/user-attachments/assets/b222e743-f334-47b0-9aa0-400074c3f80d


Piccomaster:
Ah yes, indeed I think that's enough for a single mythical weapon, the effect is great.

PR #418, opened 2026-04-16
fix some error

ATT2CNTranslationTeam:
I've updated the placement method for some items; they are now placed using display entities.


Piccomaster:
Great work!

PR #420, opened 2026-04-18
update fixed entity

ATT2CNTranslationTeam:
You need to enter `function att2:summon/fixed_entity/update` and wait for a while to update all entities (approximately). The existing NPCs in the map will be cleared. After the execution is complete, you need to enter `function att2:summon/fixed_entity/pnj_summon` to regenerate all NPCs.


ATT2CNTranslationTeam:
There's a small issue: was the block at this position placed by mistake or intentionally? Because it clearly doesn't belong to this dimension...
`/setblock -5300 69 -5107 minecraft:lime_stained_glass`
image

Piccomaster:
> There's a small issue: was the block at this position placed by mistake or intentionally? Because it clearly doesn't belong to this dimension...

Yes, that block was intentional.

I'm going to try a complete entity regeneration. I hope you're certain of the result, because the NPC entities haven't been generated or touched since the release of the very first version of the map, which was 1.0.0, 6 years ago... ^^

PR #422, opened 2026-04-21
Update character dialogue detection.

ATT2CNTranslationTeam:
Now, when a player talks to an NPC, the NPC's ID is synchronized. When the player clicks "Buy" in the chat box, it will no longer cause the issue where items displayed by Merchant A can be purchased from Merchant B.


ATT2CNTranslationTeam:
There's a small issue here: the skull generated by the `summon minecraft:armor_stand -5091 129.5 -6422.7 {Invulnerable:1b,NoBasePlate:1b,NoGravity:1b,ShowArms:1b,equipment:{feet:{id:"minecraft:leather_boots",count:1,components:{dyed_color:12286526}},legs:{id:"minecraft:leather_leggings",count:1,components:{dyed_color:7213836}},chest:{id:"minecraft:leather_chestplate",count:1,components:{dyed_color:526630}},head:{id:"minecraft:player_head",count:1,components:{profile:{properties:{name:Edi_95x}}}}},Pose:{Body:[0f,182f,0f],Head:[40f,20f,-50f],LeftLeg:[84f,176f,0f],RightLeg:[274f,0f,0f],LeftArm:[5f,0f,3f],RightArm:[354f,8f,0f]},Rotation:[0.0f,0.0f],DisabledSlots:2039583}
` command is a Steve head, but it should actually be the head of the guard at that location. Their data is different. So, should the head of this armor stand be replaced with the guard's head from that position? I'm not sure, because there might have been other variations initially...

Piccomaster:
> There's a small issue here: the skull generated by the summon command is a Steve head, but it should actually be the head of the guard at that location. Their data is different. So, should the head of this armor stand be replaced with the guard's head from that position? I'm not sure, because there might have been other variations initially...

Ah yes, that's the head of the guard killed by Korlaphe in Eolorion when he went through the Elcheol mountain pass. I'll see about changing it then.

PR #429, opened 2026-04-25
Add Chinese update log.

ATT2CNTranslationTeam:
You know what? The character count of our update log is almost reaching 30,000...


Piccomaster:
Hahaha, speaking of which. That's also why this new version won't be called "Across The Time 2 - Time for Regrets v6.0.0" but rather "Across The Time 2 - Time for Remaster v1.0.0" ^^.
The work involved is so enormous, with an almost complete overhaul of the code, that it's necessary for me to put things into perspective.

PR #431, opened 2026-04-27
update "update/all"|fix pnj summon

ATT2CNTranslationTeam:


Piccomaster:
Do I still need to run the update functions? Because I see some changes apparently, which ones do I need to run?

ATT2CNTranslationTeam:
"function att2:update/start" and "function att2:summon/fixed_entity/update" need to be executed

I also used these two commands to update based on version 5.0.1, so that I could test some basic functionality.

By the way, did you receive a message from one of my team members? He said you didn't reply, and I'm worried the message might not have gone through.

Also, I need the specific location of the NPC Eldric you added. Of course, it would be best if you could add it to the `adventquest_function\data\att2\function\summon\fixed_entity\pnj_summon.mcfunction` or other related NPC files.

Piccomaster:
Okay, so the position of the new NPC Eldric is: -5044 109 -4886 that's the position of the block under his feet; so his exact spawn position is: -5044 110 -4886.

As for message, it's normal, I didn't check the messages yesterday. They're really impatient over there, haha. Actually, one of the members seemed to think the May 1st test date would be postponed, but I didn't say anything official. I'll reassure him that they'll have an ALPHA version on May 1st.
He also said he's a mapmaker and that he could help us fix the bugs. In that case, I'll share the GitHub link with him.

ATT2CNTranslationTeam:
> Okay, so the position of the new NPC Eldric is: -5044 109 -4886 that's the position of the block under his feet; so his exact spawn position is: -5044 110 -4886.

Uhh, well, I'm not sure. It's best to point your cursor at the NPC entity and press F3+I to get the entity's coordinate data (the kind with decimals). Anyway, you can add it directly to the relevant files, since I don't have the related structures here, so there might be deviations.

One more thing — I noticed you previously moved a "Symbol" position. Did you rearrange the light source blocks as well?

By the way, I did a simple test on a map that was updated based on version 5.0.1, and there were basically no major issues.

Piccomaster:
Yes, of course, don't worry, I know the symbol system; I created it myself. ^^

Otherwise, I'm still checking all the changes you're making, and I don't understand why you removed the progress detection related to the CRT stat? Here:
`adventquest_function/data/att2/function/advancement/test_all/combat/stat.mcfunction`
All these lines deleted:
```
#CRT
execute if score @s CRT_TOT matches 2.. run advancement grant @s only att2:combat/crt_2
execute if score @s CRT_TOT matches 4.. run advancement grant @s only att2:combat/crt_4
execute if score @s CRT_TOT matches 7.. run advancement grant @s only att2:combat/crt_7
execute if score @s CRT_TOT matches 10.. run advancement grant @s only att2:combat/crt_10
```

Also you reset the total detection number of the number of exploit points here (SQ60 SIDEQUEST) to 20286:
`adventquest_function/data/att2/function/cinematic/sidequest/go.mcfunction`

Finally, once again, don't forget to update the changelog file, as the balancing changes are significant and I don't want to miss anything or be surprised by unexpected changes during testing.

PR #439, opened 2026-05-01
fix bug

ATT2CNTranslationTeam:
@Piccomaster I've organized the update log for the next update in the Google Sheet, making it easier for testers to view.
And you can also use this content to organize the update log accordingly.

image

If possible, you could choose to update the datapack and resource pack tomorrow morning.

Of course, I've already tested the fixes in my test group, so no need to worry.


Piccomaster:
Ok, I see you're still fixing things. Before doing the ALPHA 1.0.2 update, I'll take the opportunity to make a few corrections as well. Let me know when you're finished and ready to launch the update.

ATT2CNTranslationTeam:
Oh, no worries — it can be released now.

Piccomaster:
I just want to correct a few French translations that still bother me a little. I'll do that before releasing version 1.0.2.

PR #441, opened 2026-05-03
Add a new injury system.

ATT2CNTranslationTeam:
Add a new injury system:
When the player's health is too low,
Head injuries: Causes temporary blindness.
Chest injuries: No special negative effect.
Leg injuries: Prevents jumping.
Foot injury: Movement speed is reduced.
Hand injuries: Prevents melee attacks.

While an injury is active, the player continuously consumes a significant amount of hunger but also gains a certain amount of health regeneration.


Piccomaster:
Interesting, don't forget to mention it in the changelog file.

Also, pay attention to the latest fixes I made, especially those in this folder:
adventquest_function/data/att2/function/trigger/

ATT2CNTranslationTeam:
The content of this book is as follows: the correct path to call Consciousness now is items/consciousness.

PR #452, opened 2026-05-07
fix bug

ATT2CNTranslationTeam:
@Piccomaster 
Okay, I'm planning to redo all the tutorials. I think I need to delete all the images and recreate them. That should be fine, right?

The new tutorial will still use the achievement images from before, but only the images will be shown — no text will be displayed in the achievement section. The text will be moved to the chat bar or title.

Since achievement images only appear for a very short time, players don't have enough time to absorb the tutorial text. So, I will place the related information in the chat bar instead.

Okay, I feel like I might not be that good at making images. Anyway, I'll roughly put together some simple images first. If you think they're not great, you can go ahead and rework them...


Piccomaster:
I have no problem completely revising the tutorial, I didn't do any extraordinary editing either ^^.

PR #455, opened 2026-05-09
Update the new tutorial system.

ATT2CNTranslationTeam:
https://github.com/user-attachments/assets/40ea7b78-d5f1-436a-8374-986bbba3df84

I've uploaded the image data to the resource pack PR. The French part is currently AI-translated and may require further adjustment.


Piccomaster:
Just to clarify, there are physical checkpoints at each runic trial now, so there's no point in keeping an automatic checkpoint anymore.

ATT2CNTranslationTeam:
Yes, but that doesn't mean all players will update the map. So it's probably fine to trigger the respawn point at the very beginning.

Piccomaster:
Yes, you're right. Okay, that works.

PR #460, opened 2026-05-13
Update the new damage calculation system.

ATT2CNTranslationTeam:


Piccomaster:
That's a lot of changes, great work! We'll have to test everything. I also saw that you updated the Update file; I suppose I need to restart the update all process before updating to ALPHA 1.1.0 ?
However, given all the changes, would that require players to restart the map I think.

ATT2CNTranslationTeam:
There's currently no need for players to update the map — it's just a few scoreboard updates, and I've already placed them all in the load file. Players will automatically update upon entering the game. Of course, if someone is willing to start a fresh playthrough for testing, that would be even better.

PR #479, opened 2026-05-26
Optimized the detection method for spell experience.

ATT2CNTranslationTeam:


Piccomaster:
I was wondering what exactly has changed in the spell level experience?

Also, is it normal to have a "cap" score name for the last level?
Capture d&#39;écran 2026-05-26 084818
Capture d&#39;écran 2026-05-26 084808

ATT2CNTranslationTeam:
For example, directly granting 1,000 spell experience to a level 1 spell would synchronize its level to the corresponding level instantly after casting, rather than increasing its spell level one step per cast.

As for the detection of the final level — yes, that was a small mistake, and I've already fixed it.

As for the issue from yesterday — if you'd like, you can go ahead and release the new version right now.

Piccomaster:
Okay great, I think I'll send an update of a new ALPHA version tomorrow.

PR #481, opened 2026-06-11
add maze map

ATT2CNTranslationTeam:
https://github.com/user-attachments/assets/855537ef-b724-423f-af85-92b441636b20


Piccomaster:
Wow, I'm speechless... It's just amazing! I can't wait to try it out when I have the time. Working on this must have been incredibly time-consuming; I see the number of textures you added, it's enormous. Is that for every frame or for different positions each time?

ATT2CNTranslationTeam:
The principle behind this map is that I pre-processed the maze by cutting and storing it in chunks of 102 blocks wide by 51 blocks high. When loading the map, I only need the maze's starting coordinates, ending coordinates, and the player's current position to calculate the corresponding position within the data grid.

Since the title text cannot directly include line breaks, I had to use custom fonts to generate images for rows 1 through 51. Then, by using negative spacing to reverse-indent the images, I achieved the effect of line breaks. Because the maximum pixel size for custom font images is 256x256, and each cell is limited to a 5x5 pixel image, the end result was a map that could only display 51 rows.

As for how the map updates — it only refreshes when the player moves, or when a torch is placed or broken. At all other times, no detection occurs.

Performance-wise, the consumption is currently somewhat high, but I think it's still acceptable. The main reason is that the title command has to parse a very large amount of data during each map update — since the map size is 102x51, that's over 5,000 text entries being parsed each time the map refreshes. Of course, as long as you don't choose to keep the map open, there won't be excessive performance drain.

As for why I didn't make the map smaller — I felt that if the map area was too small, it wouldn't be very useful. In short, if players find the performance cost unacceptable,We can reduce the map size in a future update.

Of course, in theory, more features could still be added — like an overall map preview, or more editing functions. But I simply don't have enough time or energy to keep updating and optimizing this content. And with this kind of map feature, as long as you want, there will always be all sorts of features you could add...

Piccomaster:
Okay, I understand, that's really clever.

Regarding optimization, we'd indeed need to look at the feedback, but since it's contained separately from the rest of the map and these chunks with this system only load when the player is navigating the maze, it should be fine.

Perhaps if several players are constantly refreshing the map at the same time, it could cause problems, but I'll see about that when there's feedback later during the BETA.

Speaking of which, I think I'm going to release the BETA version of the map open to everyone, like an official release version.

Because I know I won't be able to finish my dungeon for several months, and people are starting to get impatient...
I think you also want to take a break and move on to something else, even though we'll still need to be active during the BETA to address the numerous feedback points that will arise, but the sooner the better. In short, I'll release my dungeon as soon as the beta phase ends.

What do you think would be the ideal time for an official release within the Chinese community?

ATT2CNTranslationTeam:
I don't have any special requirements regarding the release of the beta version on my end. That said, I think we could release it either this weekend, next Monday, or maybe the week after next?

As it stands, although some content is still not finished, fully completing everything would still take quite some time. Up to now, the existing content has mostly been optimized and polished.

Piccomaster:
Okay, I think I'll keep the ALPHA version running until the end of June anyway. Then, one weekend in early July, I could release the BETA version. Summer vacation is starting for many people in Europe, so young people who are out of school might be more likely to want to start playing the map.

Anyway, that works. Plus, it'll give me time to polish a few things on my end, like translations, articles, updating the website, etc.

PR #483, opened 2026-06-28
update

ATT2CNTranslationTeam:
This is a letter of temporary farewell—at least, I sincerely hope it is a temporary farewell rather than a final farewell.

As you may have observed, the time I am able to devote to development has been steadily diminishing, and the volume of content I produce has likewise declined. This is because I simply no longer have sufficient time and energy to properly attend to these endeavors. Moreover, due to various challenges in my personal life, it is possible that I will not return for several months, or perhaps even for several years. Of course, should time permit, I will continue to follow the developments within the ATT2 community; however, I will be largely unable to participate in the creation or revision of content.

I have contemplated many words, yet I find myself at a loss as to what exactly to say. ATT2 is a project that holds profound significance for me. Never did I anticipate that my life would present such a remarkable journey, nor did I ever expect that the content I created would receive such widespread support and recognition, nor did I imagine that I would be capable of accomplishing so much. If it were within my power, I would gladly dedicate every day to the creative work on ATT2; there is still so much more I wish to achieve—but reality, inevitably, will not permit me to do so.

In any case, I will return. I cannot, however, promise on which day that will be. It is even conceivable that by the time I come back, this project may no longer exist. Nevertheless, I will certainly return. I sincerely hope that all remains well.


Piccomaster:
It is with a touch of sadness that I learned this news. But indeed, projects like these can be very energy-intensive and time-consuming, especially considering how much work you've dedicated to them lately.
I know how easily, with this kind of passionate project, one can neglect oneself, one's health, or anything else related to one's personal life. And when problems arise, taking a step back is wise and the best solution.
So I won't make you feel guilty about leaving the project, even if it's only temporary; I completely understand.

I would have hoped that during the beta this summer you would be available to fix any remaining bugs, particularly those concerning the systems you developed.
But if that's impossible for you, then I won't insist. You've already contributed so much to the project; I don't even know how to thank you.

Right now, I'm feeling a bit tired and sometimes a little unmotivated, but I know the end of this map remaster project is approaching, and I have to push myself to finish it.
I'll try my best to complete it, hoping the bugs aren't too significant. But after this intense alpha test, we've already made so much progress.

You have undeniable talent, and with continued hard work, I'm sure your future projects will be a resounding success.
In any case, I hope to see you again someday, and you know you're always welcome here.

orian34:
You've been a significant contributor to this project, and that was great help.

I also took a step back this year after going through text correction but I'm hoping to free enough time for my deep dive run through the beta version after summer, where I'll be able to fix issues I encounter as I go. It'll be a good way to get more familiar with all the new systems added.

